Issue #40, opened 2015-03-05
"glide update" does not work as expected


roblillack:
This is somewhat related to #30 but got forgotten when you created single issues.

Adding a new dependency to glide is somewhat of a hassle. After adding `github.com/spf13/viper` to my `glide.yaml` this is what the output of `glide update` looks like:

```
[INFO] Updating github.com/spf13/viper with Git (From )
[INFO] Looks like github.com/spf13/viper is a new package. Cloning.
[WARN] Update failed for github.com/spf13/viper: exit status 128

…

fatal: reference is not a tree: 90734830d1ec30c19cb03e3ecd76a43dc1618b01

[WARN] Failed to set version on github.com/spf13/viper to 90734830d1ec30c19cb03e3ecd76a43dc1618b01: exit status 128

…

[INFO] Running go build github.com/spf13/viper
[WARN] Failed to run 'go install' for github.com/spf13/viper: can't load package: package github.com/spf13/viper: cannot find package "github.com/spf13/viper" in any of:
        /usr/local/Cellar/go/1.4.1/libexec/src/github.com/spf13/viper (from $GOROOT)
        /Users/rob/dev/darling/_vendor/src/github.com/spf13/viper (from $GOPATH)
```

Three errors here:
* Update should resort to installing this dependency. Glide notices this but still fails.
* This error shall be reflected in glide's exit code.
* Ideally, glide stops at the first error.

If I run `glide install` I get lot of warnings about all those other dependencies which are up-to-date. spf13/viper is installed correctly but cannot be built due to missing dependencies. Again, I'd expect glide to stop here, but it will build all other dependencies so I have to scan the whole output for errors.

glide version 0.2.0-43-g1a5c44e

roblillack:
Pull request #41 fixes the first point, #42 the third.

mattfarina:
@roblillack thanks for providing this. I'll take a look at it soon.

technosophos:
I think we can mark this closed now, right @mattfarina ?

mattfarina:
I think so. If there is a continued issue that's not been address or isn't in one of the other issues this can be re-opened.

PR #41, opened 2015-03-05
Fail early when unable to update or build dependencies.

roblillack:
This patch implements a "fail early" strategy in some crucial places.

Please see issue #40 for more information.

### DIFF
--- a/cmd/get_imports.go
+++ b/cmd/get_imports.go
@@ -45,7 +45,7 @@ func UpdateImports(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Inte
 
 	for _, dep := range cfg.Imports {
 		if err := VcsUpdate(dep); err != nil {
-			Warn("Update failed for %s: %s\n", dep.Name, err)
+			Error("Update failed for %s: %s\n", dep.Name, err)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


mattfarina:
I spoke with @technosophos and we need to think this one over. Since I spoke with him I've wondered about a couple different issues here. Wanted to let you know we had not forgotten about this.

roblillack:
What are those issues? I'd be happy to discuss here!

mattfarina:
@roblillack I'd like @technosophos to speak up as well. I'm wondering if displaying an Error and exiting like this is a good idea state wise. Two things come to mind.

First, should we use a two pass system. On the first pass glide looks for things that will lead to errors, report them, and then not try the action. The idea is to catch something that could put the environment into a bad state and stop it before it happens.

Second, I don't like that it's easy to leave _vendor is a bad state. Both prior to this change and after the change. I'm not sure of the best approach, other than a 2 pass system, but I'd like to explore it.

Any ideas?

technosophos:
If the VCS provider goes down, and we can't get an update, will this stop builds? It looks like it will. And that's my only hesitation. Say I have one package on an unreliable Git server. I'd rather not have it fail just because that server went down. My CI system would hate that.

Otherwise, I think I am good with @roblillack 's patch.

I'm not sure this patch makes the state of _vendor any worse than it already is. So I wouldn't hesitate on that front. And I think that rather than using a "two-pass" system, I'd ultimately rather go to a rollback-on-fail if that's feasible. But that's way beyond the scope for this request.

roblillack:
I don't think, merely building the software shall use a network connection at all -- see pull request #45 for a Git-specific fix. Having non-working builds, was the reason I investigated this issue in the first place: On our CI system, random issues trying to set up dependencies would fail without stopping the whole build, which in turn leads to a whole bunch of different errors way later in the build process.

technosophos:
I'm thinking we should give this patch a try if we go ahead with #45. Are you okay with that, @mattfarina ?

mattfarina:
After talking this over with @technosophos we're going to pass on merging this. It's quite out of date given all the changes recently and most things are warnings now where we don't want to exit.

PR #42, opened 2015-03-05
Fix 'glide update' for new dependencies, if no repo configured for git/hg.

roblillack:
For simple dependency configurations like:

```
  - package: github.com/spf13/viper
    vcs: git
    ref: 90734830d1ec30c19cb03e3ecd76a43dc1618b01
```

(package/vcs specified, but no repo) `glide update` did not work as expected. This patch brings the update feature up on par with `glide install` for these cases.

Please also see issue #40.


mattfarina:
Reviewed and tested. Thanks for the fix.

Issue #30, opened 2014-10-23
Semantic of install/update commands unclear


roblillack:
Hi,

the terminology of the two commands _install_ and _update_ ist somewhat misleading to me:

- Why is there a difference at all? If _update_ also installs dependencies, why have another command for that?
- If I change the `repo` of one of the packages in glide, I expect _update_ to:
  - Check, if there are local changes to the package, and FAIL with an error message.
  - If only the ref was changed, try to use whatever methods the VCS provides to check out that new ref.
  - If repo or VCS was changed, completely clear the src directory and re-fetch it using the configured VCS and repo.

  I _think_ that right now, only the second part is done, but the documentation is not really clear about this. Also, does it only do something along the lines of `git pull` or, does it check `glide.yaml` for changes, too?

mattfarina:
I'd like to close this issue out. Most of the elements have their own issues carrying forward. In 0.5.0 `install` and `update` are commands aliased to each other. It's a matter of semantics at that point.

If there is some unresolved issue not documented somewhere please re-open or create a new issue.

PR #44, opened 2015-06-01
Add support for changed VCS type or repository for Git. Issue #30.

roblillack:
This patch adds the much-needed functionality of checking the repository type and remote of the current checkout of a git dependency against the configuration.

If the remote `origin` does not match whatever is configured as repo, a fresh clone is initiated after wiping the dependencies source directory.

As discussed in #30, this allows switching to a fork, or different remote location for any git dependency, by merely changing `glide.yaml` (and running `glide update` for example as part of the CI process).

### DIFF
--- a/cmd/git.go
+++ b/cmd/git.go
@@ -14,7 +14,6 @@ type GitVCS struct{}
 var _ VCS = &GitVCS{}
 
 var (
-	NoWorkingDirectory  error = errors.New("Working directory does not exist")
 	WrongVCS            error = errors.New("Wrong VCS detected")
 	CannotDetermineRepo error = errors.New("Unable to determine repository")
 )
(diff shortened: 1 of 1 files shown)


technosophos:
YES! YAY! Excellent patch! This fixes a major limitation of glide. I am so excited to see this!

@mattfarina If you okay too, I will merge.

mattfarina:
This is great. Needs a tweak before I want to merge.

Note, this only works if the switch is to git. If the repo type changes from git to hg this doesn't work. I'm ok with that for now but there must be a way to add this to other VCS more cleanly.

cmd/git.go
```diff
@@ -10,6 +11,32 @@ import (
...
+	WrongVCS            error = errors.New("Wrong VCS detected")
+	CannotDetermineRepo error = errors.New("Unable to determine repository")
+)
+
+var remoteRegex = regexp.MustCompile("^origin\\s+(\\S+)\\s+\\S+$")
+
+// returns the currently checked out remote repository
+// according to the state of the working directory
+func (g *GitVCS) currentRepository() (string, error) {
+	out, err := exec.Command("git", "remote", "-v").CombinedOutput()
```
mattfarina: I checked out cookoo as an hg repo in the right place in the `_vendor` directory. Then I ran `glide update` from the root of the glide project. I saw it noted:
```
[INFO] Repository changed from git@github.com:Masterminds/glide.git to git@github.com:Masterminds/cookoo.git. I'm doing a clean checkout.
```
The wrong repo was listed as the one it's coming from. When `git remote -v` is run from the hg repo it bubbles up all the way to glide itself and provides that info rather than staying scoped to the _vendor directory.

For VCS changes this is wonkey. Any ideas how we could better handle this?

mattfarina:
I have a fix for the directory bubbling up so I'll merge this and then add the fix. Thanks for bringing this up @roblillack.

PR #45, opened 2015-06-01
Skip "git fetch" if current checkout matches configured ref or tag.

roblillack:
Upon updating a Git dependency, check if the current checkout matches whatever _ref_ or _tag_ (intentionally skipping branches that might change on the remote side, tags SHOULD not) is configured.

This speeds up the `glide update` run tremendously for bigger projects.

### DIFF
--- a/cmd/git.go
+++ b/cmd/git.go
@@ -10,6 +10,30 @@ import (
 
 type GitVCS struct{}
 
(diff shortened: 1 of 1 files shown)


technosophos:
I'm definitely good with this patch. @mattfarina if you're okay with it, I'll merge.

mattfarina:
@roblillack Nice job. Tested and it works great.

Issue #55, opened 2015-08-21
PROPOSAL: Add ability to declare how to deal with multiple copies of the same package


technosophos:
With Glide 0.5, it is now possible for two packages to declare the same dependency. During deep imports (`glide up` and `glide get`), each time a dependency is found, it is stored in the `vendor/` directory closest to the dependency.

So you can end up with cases where the same dependency appears multiple times:

```
vendor/foo/bar
vendor/x/y/z/vendor/foo/bar
```

This naively resolves the case of package collisions. However, there are some clear cases where this is not desired. Namely, where having one code base using multiple versions will lead to bugs.

## Proposed solution:

It seems that that the best place to handle this is in the top-level `glide.yaml`. I propose adding the following directives to the `glide.yaml` `package` object:

```
- package: foo
  required: once
- package: bar
  required: never
- package: baz
  required: any
```

Here's the behavior of `required`:

* once: Only one copy of this library is ever checked out, and the version is set to the version in the topmost `glide.yaml`. Thus, the combination of `once` plus `ref` will effectively pin the package on any sub repositories. The definitive copy will be stored in the top-level `vendor/`.
* never: Regardless of what Glide finds in the dependency tree, don't ever import a copy. This is a good idea to add, since we want to play nicely with other package management tools and techniques, which may need to manage a dependency on their own.
* any: The default strategy. Get any number of copies of the dependency. Store as close to the import as possible. This will result in deeply nested repos.

abourget:
I also think this is needed.. however the semantics don't stand out as being ultra clear to me.. especially `require: never` and `require: any`.. someone not having this proposal would have a hard time knowing exactly what to expect out of this.

I'm trying to think of other ways to name this.. names stick.. so let's choose well.

technosophos:
How about something like `allowMultipleCopies` with values `yes`, and `no`. For the `never` case, maybe we should handle that separately.

abourget:
`canonical: true`, `singlecopy: true`, `topmost: true`, `resolve: canonical`, `once: true`, `no_copy: true`, `no_checkout: true`, 

I think the `never` isn't related to the fact we're doing a single import, or multiple imports, but rather the fact that we're doing an import at all.. so something like `no_import: true`, `ref_only: true`, `reference: true`.. but that doesn't sit well with the goal of a `glide.yaml` which is to list deps.. not sure about this one

[...]

Issue #52, opened 2015-08-19
support semantic versioning of dependencies a la gopkg


azylman:
I'd love if the `glide.yml` file could specify semantic versions, so I could specify something like "version 1" and then it would pull the most recent `1.x.x` tag off of github, similar to what gopkg does

mattfarina:
Yes! As a user of tools like npm and composer I know exactly what you mean. This seems like a good candidate for the 0.6.0 release.

mattfarina:
I'm looking at two methods for version comparisons. One is strict to SemVer and the other is looser. I blogged about them recently at http://engineeredweb.com/blog/2015/go-version-compare/

azylman:
fwiw, I've used [gopkg.in/blang/semver.v1](http://godoc.org/gopkg.in/blang/semver.v1) before and was happy with it (used it in [gitsem](https://github.com/clever/gitsem), which we use to version all of our libraries)

technosophos:
I think that's a good idea. I would suggest that we support similar rules to gopkg.in, but with more precision:

v1 matches v1, v1.x, v1.x.y

v1.1 matches v1.1 and v1.1.y

v1.1.1 matches v1.1.1

And since one version of semver requires the v, while the other does not, we'd want to support the version without v, too.

mattfarina:
For anyone wanting to follow along, the first step is to support branches and tags on [Masterminds/vcs](https://github.com/Masterminds/vcs). I started the branch feature/references to begin that work (and it's started). It will take me at least a few days to work through the best way to do this for each of the VCS.

mattfarina:
Started the work on the feature/semver branch.

azylman:
Can't wait to see this!

mattfarina:
To keep the details up to speed for anyone wondering, there is an initial version on the _feature/semver_ branch. But, we're not all that keen on the constraints we can use and how this is working. I've been trying some existing semver libraries.

Due to the lack in features I've started a [semver](https://github.com/Masterminds/semver) library. The goal is to have many of the features you get in similar libraries in other languages. Once this is in good shape (in maybe a week) I'll update the _feature/semver_ branch and we'll look at merging.

This is the big thing I'm working on at the moment for Glide.

mattfarina:
#114 (a pull request from a feature branch) added SemVer support. Still needs to be documented but it's there.

mattfarina:
With the basic docs being present I'm going to close this. Happy SemVer'ing.

[...]

PR #70, opened 2015-08-30
Add flatten open to imports

interlock:
This is a prototype of #55 

It adds a `flatten` option that takes `true` or `false` as an option.

When set to `true`, any sub package vendored in with the same name will be skipped.
When set to `false`, business as usual.

A few other changes of note:
- []*Dependency was made into a type Dependencies so I could put common `Get` by name functionality on it. There may be code elsewhere that can use this.
* Added `boolOrDefault` to work with the yaml and having a default bool set
* Added the concept of a Config.Parent and some funs related to it traversing up the tree.
* Enhanced some of the yaml tests for new and existing functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,6 +325,10 @@ Darwin/OSX systems:
 
 The package will not be fetched for other architectures or OSes.
 
+**Q: How do I prevent vendored packages from importing the same package**
+
+You can use the `flatten: true` config option on the entire project or just one specific dependency.
+
 ## LICENSE
 
 This package is made available under an MIT-style license. See
--- a/cmd/get_imports.go
+++ b/cmd/get_imports.go
@@ -193,7 +193,7 @@ func VcsExists(dep *Dependency, dest string) bool {
 	if err != nil {
 		return false
 	}
-	Info("Repo not found %s, checking local\n", dep.Name)
+
 	return repo.CheckLocal()
 }
 
@@ -239,7 +239,6 @@ func VcsUpdate(dep *Dependency, vend string, force bool) error {
 		if empty == false && err == v.ErrCannotDetectVCS {
 			Warn("%s appears to be a vendored package. Unable to update. Consider the '--update-vendored' flag.\n", dep.Name)
 		} else {
-			Info("About to GetRepo %s\n", dest)
 			repo, err := dep.GetRepo(dest)
 
 			// Tried to checkout a repo to a path that does not work. Either the
--- a/cmd/get_imports.go
+++ b/cmd/get_imports.go
@@ -193,7 +193,7 @@ func VcsExists(dep *Dependency, dest string) bool {
 	if err != nil {
 		return false
 	}
-
+	Info("Repo not found %s, checking local\n", dep.Name)
 	return repo.CheckLocal()
 }
 
@@ -239,7 +239,7 @@ func VcsUpdate(dep *Dependency, vend string, force bool) error {
 		if empty == false && err == v.ErrCannotDetectVCS {
 			Warn("%s appears to be a vendored package. Unable to update. Consider the '--update-vendored' flag.\n", dep.Name)
 		} else {
-
+			Info("About to GetRepo %s\n", dest)
 			repo, err := dep.GetRepo(dest)
 
 			// Tried to checkout a repo to a path that does not work. Either the
--- a/cmd/recursive_glide.go
+++ b/cmd/recursive_glide.go
@@ -96,17 +96,18 @@ func dependencyGlideUp(parentConf *Config, base string, godep, gpm, force, delet
 		wd := path.Join(vdir, imp.Name)
 		// if our root glide.yaml says to flatten this, we skip it
 		if dep := conf.GetRoot().Imports.Get(imp.Name); dep != nil {
-			if dep.Flatten == true {
+			flatten := conf.GetRoot().Flatten
+			if flatten == true && dep.Flatten == false ||
+				flatten == false && dep.Flatten == true {
+				flatten = dep.Flatten
+			}
+			if flatten == true {
 				Info("Skipping importing %s due to flatten being set in root import glide.yaml\n", imp.Name)
-				imp.Flattened = true
-
-			} else if conf.GetRoot().Flatten == true {
-				Info("Skipping importing %s due to flatten being set in root config glide.yaml\n", imp.Name)
 				imp.Flattened = true
 			}
 
-			if imp.Reference != dep.Reference {
-				Warn("Main vendored package %s ref (%s) is diferent from sub vendored package ref (%s)\n", imp.Name, imp.Reference, dep.Reference)
+			if flatten == true && imp.Reference != dep.Reference {
+				Warn("Flattened package %s ref (%s) is diferent from sub vendored package ref (%s)\n", imp.Name, imp.Reference, dep.Reference)
 			}
 
 			if imp.Flattened == true && deleteFlatten == true {
--- a/cmd/recursive_glide.go
+++ b/cmd/recursive_glide.go
@@ -105,6 +105,10 @@ func dependencyGlideUp(parentConf *Config, base string, godep, gpm, force, delet
 				imp.Flattened = true
 			}
 
+			if imp.Reference != dep.Reference {
+				Warn("Main vendored package %s ref (%s) is diferent from sub vendored package ref (%s)\n", imp.Name, imp.Reference, dep.Reference)
+			}
+
 			if imp.Flattened == true && deleteFlatten == true {
 				if exists, _ := fileExist(wd); exists == true || true {
 					remove := wd + string(os.PathSeparator)
--- a/cmd/recursive_glide.go
+++ b/cmd/recursive_glide.go
@@ -40,13 +40,14 @@ func recDepResolve(conf *Config, vend string, godeps, gpm, force bool) (interfac
 	if len(conf.Imports) == 0 {
 		Info("No imports.\n")
 	}
-	Info("Config: %+v\n", conf.Imports[0].Name)
 
 	// Look in each package to see whether it has a glide.yaml, and no vendor/
 	for _, imp := range conf.Imports {
+		if imp.Flattened == true {
+			continue
+		}
 		base := path.Join(vend, imp.Name)
 		Info("Looking in %s for a glide.yaml file.\n", base)
-		Info("\tFlatten: %b", imp.Flatten)
 		if !needsGlideUp(base) {
 			if godeps {
 				importGodep(base, imp.Name)
@@ -59,7 +60,8 @@ func recDepResolve(conf *Config, vend string, godeps, gpm, force bool) (interfac
 				continue
 			}
 		}
-		if err := dependencyGlideUp(base, godeps, gpm, force, imp.Flatten); err != nil {
+
+		if err := dependencyGlideUp(conf, base, godeps, gpm, force); err != nil {
 			Warn("Failed to update dependency %s: %s", imp.Name, err)
 		}
 	}
@@ -68,7 +70,8 @@ func recDepResolve(conf *Config, vend string, godeps, gpm, force bool) (interfac
 	return nil, nil
 }
 
-func dependencyGlideUp(base string, godep, gpm, force bool, flatten bool) error {
+func dependencyGlideUp(parentConf *Config, base string, godep, gpm, force bool) error {
+	Info("Doing a glide in %s\n", base)
 	//conf := new(Config)
 	fname := path.Join(base, "glide.yaml")
 	f, err := yaml.ReadFile(fname)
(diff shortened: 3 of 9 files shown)


technosophos:
This looks great! As I understand the patch, flattening will always favor the version that is declared in the top-level 'glide.yaml'. So if the top pins on version `1.1.1` and another dependency pins on `1.1.0`, the top level will "win", right?

Do you think we ought to allow setting a default Flatten behavior at the `Config` level, and then be able to override it in subsequent `Dependency` sections?

interlock:
Thanks @technosophos 

Yes, right now it flattens based on the root glide.yaml. I think to do it any other way would require traversing the tree, which is not complete until you load everything.

I do wonder if it makes sense to emit a warning if the root `ref` is different from one lower in the tree? I could see some frustrating bugs coming out of not having any notification that a flatten was forcing an incompatible version.

It would be fairly easy to make flatten a `Config` level setting as well. That was actually my first iteration, but I factored it out in favour of being specific.

technosophos:
I think emitting a warning would be good if mismatched refs were found. @mattfarina may have some ideas on how the VCS code might tackle that in a more sophisticated way.

mattfarina:
I agree with @technosophos about emitting a message when the versions are different. That will help developers see the issues. This will be useful when we get to SemVer #52 (which is being discussed as a recommendation for Go packages, too).

We may need a common way to get tags/branch names for comparison purposes. That would need to live in https://github.com/Masterminds/vcs and work for all the VCS. There are clues in `golang.org/x/tools/go/vcs` to how we could do this.

interlock:
@mattfarina Any reason a strict comparison of Dependency.Ref would not do for now? 

Once semver comes another update will need to be made to compare versions with fuzz.

interlock:
Does it make sense to add removing a sub vendor package if it see's it marked `flatten` and it exists? Otherwise, adding `flatten` after the first install or update would require it to be manually removed.

technosophos:
I think we should totally do strict comparison for now. But I think we will move to something more sophisticated in the future.

For example, if a top level says `v1` and a dependency says `v1.2.2`, we might choose the dependency's version (in other words, specificity wins). But that we can *definitely* do later.

We've learned from experience that people prefer to explicitly specify when things should be deleted. Should we rely on the existing `--delete` flag?

https://github.com/Masterminds/glide/blob/master/glide.go#L278

interlock:
Sound good. I will work in a Ref check that triggers a warning.

Does it make sense to overload --delete in this case or create a new specific one --delete-flatten?

technosophos:
That is a good question. Let's start safe and go with `--delete-flatten`.

I'm really excited about this feature. Thank you!

mattfarina:
Sorry for any confusion I caused. Any magic checking should go in later. Basically, what @technosophos said.

interlock:
No problem @mattfarina 

I've got delete-flatten going
Working on the simple brute Ref compare to generate a warning next.
I'll update the docs, push the updated PR and remove the WIP.

interlock:
Ready for review @mattfarina @technosophos 

* Added a dumb Reference check that Warn()'s if they differ.
* `--delete-flatten`is now a thing.
* `flatten` on the root yaml makes all the things flat, if you are into that.

hilyjiang:
Good job, @interlock 

I just check our your fork:
`git clone https://github.com/interlock/glide.git -b flatten`

And I have set `flatten: true` in the root yaml config, but `glide up` didn't import all packages in `vendor/`, my config like below:
```
package: main
flatten: true
import:
....
```

Is there anything wrong with my config?

interlock:
@hilyjiang Believe you may be actually compiling parts of the regular glide into the project the way you have it checked out.

You will want to follow steps that look something like this: 

1. $GOROOT/src/github.com/Masterminds/glide
2. git add remote interlock https://github.com/interlock/glide.git
3. git fetch interlock
4. git checkout interlock/flatten -b flatten
5. go install -a ./...

In the mean time, I'll functionally test that one more time to confirm it works :-)

hilyjiang:
@interlock I have remove all old work copy and follow your installation.

You can see that `glide` already have flatten option.
```
# glide up --help | grep flatten
        If the '--delete-flatten' flag is present, Glide will remove any depenedencies
        markred flatten within dependencies.
   --delete-flatten             Delete flattened vendor packages.
```

But it still not work as my expect.

This is a minimal test project:
 - `c` depends on `a` and `b`
 - `b` depends on `a`

```
[root@localhost ~]# cd /root/godev/src/c/
[root@localhost c]# glide up
[INFO] Fetching updates for a.
[INFO] Fetching updates for b.
[INFO] Checking dependencies for updates. Godeps: false, GPM: false
[INFO] Inspecting /root/godev/src/c/vendor.
[INFO] Looking in /root/godev/src/c/vendor/a for a glide.yaml file.
[INFO] Package a manages its own dependencies.
[INFO] Looking in /root/godev/src/c/vendor/b for a glide.yaml file.
[INFO] Doing a glide in /root/godev/src/c/vendor/b
[INFO] Importing a to project /root/godev/src/c/vendor/b
[INFO] Inspecting /root/godev/src/c/vendor/b/vendor.
[INFO] Looking in /root/godev/src/c/vendor/b/vendor/a for a glide.yaml file.
[INFO] Package a manages its own dependencies.
[root@localhost c]# tree 
.
├── glide.yaml
├── main.go
└── vendor
    ├── a
    │   └── a.go
    └── b
        ├── b.go
        ├── glide.yaml
        └── vendor
            └── a
                └── a.go

5 directories, 6 files
```

There are still two copies of `a` in the vendor directory of `c`.

hilyjiang:
@interlock I have remove all old work copy and follow your installation.

You can see that `glide` already have flatten option.
```
# glide up --help | grep flatten
        If the '--delete-flatten' flag is present, Glide will remove any depenedencies
        markred flatten within dependencies.
   --delete-flatten             Delete flattened vendor packages.
```

But it still not work as my expect.

This is a minimal test project:
 - `c` depends on `a` and `b`
 - `b` depends on `a`

```
[root@localhost ~]# cd /root/godev/src/c/
[root@localhost c]# cat glide.yaml 
package: c
flatten: true
import:
  - package: a
    repo: /git/a.git
    vcs: git
  - package: b
    repo: /git/b.git
    vcs: git
[root@localhost c]# glide up
[INFO] Fetching updates for a.
[INFO] Fetching updates for b.
[INFO] Checking dependencies for updates. Godeps: false, GPM: false
[INFO] Inspecting /root/godev/src/c/vendor.
[INFO] Looking in /root/godev/src/c/vendor/a for a glide.yaml file.
[INFO] Package a manages its own dependencies.
[INFO] Looking in /root/godev/src/c/vendor/b for a glide.yaml file.
[INFO] Doing a glide in /root/godev/src/c/vendor/b
[INFO] Skipping importing a due to flatten being set in root config glide.yaml
[INFO] Inspecting /root/godev/src/c/vendor/b/vendor.
[root@localhost c]# tree 
.
├── glide.yaml
├── main.go
└── vendor
    ├── a
    │   └── a.go
    └── b
        ├── b.go
        └── glide.yaml

3 directories, 5 files
```

It works!

Then let's go ahead:
 - `d` depends on `c`
 - `c` depends on `a` and `b`
 - `b` depends on `a`

```
[root@localhost c]# cd ../d/
[root@localhost d]# cat glide.yaml 
package: d
flatten: true
import:
  - package: c
    repo: /git/c.git
    vcs: git
[root@localhost d]# glide  up
[INFO] Fetching updates for c.
[INFO] Checking dependencies for updates. Godeps: false, GPM: false
[INFO] Inspecting /root/godev/src/d/vendor.
[INFO] Looking in /root/godev/src/d/vendor/c for a glide.yaml file.
[INFO] Doing a glide in /root/godev/src/d/vendor/c
[INFO] Importing a to project /root/godev/src/d/vendor/c
[INFO] Importing b to project /root/godev/src/d/vendor/c
[INFO] Inspecting /root/godev/src/d/vendor/c/vendor.
[INFO] Looking in /root/godev/src/d/vendor/c/vendor/a for a glide.yaml file.
[INFO] Package a manages its own dependencies.
[INFO] Looking in /root/godev/src/d/vendor/c/vendor/b for a glide.yaml file.
[INFO] Doing a glide in /root/godev/src/d/vendor/c/vendor/b
[INFO] Importing a to project /root/godev/src/d/vendor/c/vendor/b
[INFO] Inspecting /root/godev/src/d/vendor/c/vendor/b/vendor.
[INFO] Looking in /root/godev/src/d/vendor/c/vendor/b/vendor/a for a glide.yaml file.
[INFO] Package a manages its own dependencies.
[root@localhost d]# tree 
.
├── glide.yaml
└── vendor
    └── c
        ├── c.go
        ├── glide.yaml
        └── vendor
            ├── a
            │   └── a.go
            └── b
                ├── b.go
                ├── glide.yaml
                └── vendor
                    └── a
                        └── a.go

7 directories, 7 files
```

There are still two copies of `a` in the vendor directory of `c`.

interlock:
@hilyjiang I believe in #55 it is mentioned that flatten can presently only flatten imports in the root yaml file, since we do not have a way to build a graph of the imports to flatten further down.

Your second test confirms that is the case. If you need to prevent a library in a vendored package from being loaded many times, include it in the root yaml. This lets you control the Reference, which could help resolving any version issues that come up. In your second test you would import package a in your root to prevent the two copies.

Ideally we could build a import graph and figure it out, but I think golang needs to allow rewriting pkg namespaces to make that useful/practical.

interlock:
I lied, @hilyjiang, it was in this thread I mention only imports in the root will get flattened, the global flatten just implies each import has flatten. Perhaps a doc update would make that more clear? Suggestions?

hilyjiang:
@interlock Why not place all nested dependence directly under root's `vendor/` directory?

Will it cause building error or else?
Actually they are just vendors for the root package.
I don't think we should keep the nested structure for vendor packages.

For practice, I think nobody want to write `glide.yaml` with long import list to flatten, right?

interlock:
At least for our usecase there at still duplicate pkg imports in our tree that we do not want to flatten. So having import level control is important.




I am definitely open to changing how flatten works at the root glide level. A test with our code bases shows is breaks a more than it fixes, but your mileage may vary. 




I do feel making that a param on update could cause people a lot of pain. Perhaps not much harm, since they can delete and redo. At least if you have to put it in be glide file you probably understand what it implies. 



—

hilyjiang:
@interlock I agree with you that put it in config file would be better.
And for some case, import level control is really required (e.g. for different refs).

In my example above, although `c` marked as flatten, but it not actually flatten in nested `glide up`.
Only the root flatten take affected:
```
[root@localhost d]# cat ../c/glide.yaml 
package: c
flatten: true
import:
  - package: a
    repo: /git/a.git
    vcs: git
  - package: b
    repo: /git/b.git
    vcs: git
[root@localhost d]# tree 
.
├── d.go
├── glide.yaml
└── vendor
    └── c
        ├── c.go
        ├── glide.yaml
        └── vendor
            ├── a
            │   └── a.go
            └── b
                ├── b.go
                ├── glide.yaml
                └── vendor
                    └── a
                        └── a.go
```
`glide up` individually in `c` not act the same to the recursive `glide up`.

Can we change the config value of `flatten` to `root`, `self` and `off`?
- `root`: flatten all vendor packages marked with `flatten` into root vendor directory
- `self`: flatten vendor packages into current package's vendor directory
- `off`: disable flatten

interlock:
Curious what @mattfarina and @technosophos have to say. 

Feels like flatten is scope creeping on us :-)

mattfarina:
I'm going to try to find some time and give this a good review today (I hope).

What we need is a base set of functionality. This should not try to solve all the problems here. Once a sound base is in we can iterate.

mattfarina:
@interlock thanks for all your work on this.

A couple things, and @technosophos can feel free to disagree with me here:

1. I don't expect this pull request to be perfect. I see numerous place to _fix_ non-ideal code or to add features. I'm happy to work on those myself in follow-up commits or to accept pull requests for those.
2. I'm not sure about the name `--delete-flatten`. Should it be just `--flatten`?

I'd like to wait for @technosophos to provide feedback before merging this.

technosophos:
I am with @mattfarina on the idea that I'd rather merge some basic flattening support, collect some feedback from the community, and then see what the best behavior is when it comes to recursive behavior. We'll really need to think about the implications of recursive flattening. I agree with @hilyjiang that in some cases it seems like expected behavior, but I am worried that we may end up with some really subtle bugs creeping in when a top level package makes decisions about dependencies of dependencies (of dependencies....).

I think we should leave the flag as `--delete-flatten` because the idea is that it indicates that files will be deleted when the flag is added. We can flatten without deleting.

interlock:
@mattfarina `--delete-flatten` actually deletes packages that are not flattened but should now be flattened. It doesn't actually flatten (or delete) anything unless the `glide.yaml` indicates it should. 

Its use case would be you were not flattening any pkg's, updated your glide and wanted to remove any packages that should now be flat. This fixes the case where you had a pkg conflict and want to pull everyone down to the root vendor version.

@technosophos Agreed with the flatten to root vendors being a difficult one to solve. Looking at NPM, we would have to do fetch and cache everything. Process the glide's and implied glides. Resolving the graph and flattening what we could and could not. It is the most active part of their code base and quite complex. Ultimately, NPM just tries to guess what works globally with semver and moves on. At this point, glide is basic enough that we use a human to do it. I think that is a fair compromise for now.

I'll confirm that pkg level flatten overrides global flatten as mentioned in #73

interlock:
Last push adds imported pkg flatten overriding global flatten. Updated some logic around the Reference differing warning as well.

I think this is good to merge at this point.

technosophos:
I just finished a code review, and at this point I think we should merge it. Even if we decide to build something more sophisticated in the future, this is definitely the foundation we'd start with.

I'll leave it to @mattfarina to give his sign-off and merge it.

mattfarina:
@technosophos @interlock I'll merge this on Tuesday. Given that people pull directly from the tip of master to use Glide and we're about to step into a holiday weekend in the US (where I will be offline) I'd prefer to wait until then. It just seems safer.

interlock:
Many thanks from myself and the team at Picatic for this. I am available for bugs/support related to this issue.

mattfarina:
@interlock thanks for the feature. This will be useful.

PR #75, opened 2015-09-09
Fixes incorrect build context when $GOROOT env var not set

### LINKED ISSUE #74: "glide guess" is broken with Go 1.5 official package installer
### Steps to reproduce

1. Remove previous Go version and any `$GOROOT` declarations in .bash_profile/.zshrc
2. Install go1.5.darwin-amd64.pkg from golang.org
3. Switch to any project folder within the `$GOPATH`
4. Run `glide create`
5. Run `glide guess`

### Expected result
`glide guess` does not fail

### Actual result
`glide guess` fails with error:

```sh
Oops! cannot find package "fmt" in any of:
	/usr/local/Cellar/go/1.5/libexec/src/fmt (from $GOROOT)
	/Users/Janek/go/src/fmt (from $GOPATH)
```

### Additional information

`/usr/local/Cellar` does not exist on my system. It seems to be the path were Go would be installed using `brew`.

`$GOROOT` env variable is **not** set on my system. This is not necessary with the package installer, since it installs Go to the default location used by the Go binary.

The Go binary correctly uses this location as its root path:

```sh
$ go env
(...)
GOPATH="/Users/Janek/go"
GOROOT="/usr/local/go"
```

It looks like `glide` always falls back to the path where `brew` installs Go. Obviously, when Go is not installed using `brew` this will break things.

**Proposed solution:**
Parse GOROOT from `go env` output.

janeczku:
Fixes #74 and #76

### DIFF
--- a/cmd/guess_deps.go
+++ b/cmd/guess_deps.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"github.com/Masterminds/cookoo"
-	"go/build"
 	"os"
 	"strings"
 )
@@ -12,9 +11,13 @@ import (
 // Params
 // 	- dirname (string): Directory to use as the base. Default: "."
 func GuessDeps(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	buildContext, err := GetBuildContext()
+	if err != nil {
+		return nil, err
+	}
 	base := p.Get("dirname", ".").(string)
 	deps := make(map[string]bool)
-	err := findDeps(deps, base, "")
+	err = findDeps(buildContext, deps, base, "")
 	deps = compactDeps(deps)
 	delete(deps, base)
 	if err != nil {
@@ -24,7 +27,7 @@ func GuessDeps(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrup
 	config := new(Config)
 
 	// Get the name of the top level package
-	config.Name = guessPackageName(base)
+	config.Name = guessPackageName(buildContext, base)
 	config.Imports = make([]*Dependency, len(deps))
 	i := 0
 	for pa := range deps {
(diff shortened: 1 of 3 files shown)


technosophos:
I'm good with this. I am sorta in favor of removing the conditional and adding `buildContext` to the context so that any command that needs it can access it. `guess`, `tree`, and `list` definitely need it. I think `get` will likely need it in the near future.

mattfarina:
@janeczku thanks for starting this.

I'd like to see a tweak to this. There are lots of times where `$GOROOT` isn't set and it's not required to be set. We can do something more intelligent.

Instead, it'd like to see a context builder as a utility function. It could be similar to the [`defaultContext`](https://github.com/golang/go/blob/go1.5.1/src/go/build/build.go#L287) function from the `go/build` package. Instead of getting the `$GOROOT` and other environment variables from the `runtime` package it could use the `go env [VAR]` command to get them (e.g., `go env GOROOT`). Or, if it's faster just parse the output of `go env` (without the variable name) to get the needed information.

The reason I suggest it as a utility function rather than on every run of Glide is because of the calls to the `go` program. Limit those to only the times it's needed. Essentially lazy load the information.

I filed golang/go#12567 to see if we can get easier access to the data. But, that wouldn't be for awhile and we can tackle this now.

technosophos:
I was thinking about the utility function approach, too. It's slightly more consistent with what Go does, and what we've done elsewhere in our codebase. But the dependency injection style is the way we usually use Cookoo. In the present usage, though, I'm not sure we're gaining much by using DI.

I think that @mattfarina is right -- we should refactor this as a utility function to be called on demand.

janeczku:
Great feedback! I might have done this as utility function in the first place, but i got distracted by what Cookoo had to offer :) I will refactor and then update this patch probably tomorrow.

janeczku:
> Instead of getting the $GOROOT and other environment variables from the runtime package it could use the go env [VAR] command to get them (e.g., go env GOROOT). Or, if it's faster just parse the output of go env (without the variable name) to get the needed information.

It already does that :)

mattfarina:
I'm reviewing now and testing some situations out (e.g., Windows)

technosophos:
Huge thanks, @janeczku !

janeczku:
Thanks back for this awesome tool! @technosophos @mattfarina

PR #89, opened 2015-09-25
More robust GOPATH retrieval

gcmt:
This patch keeps glide from breaking if GOPATH is not set.

**EDIT**: With the second commit I essentially made both functions `Gopath()` and `Gopaths()` behave in a more predictable way.

### DIFF
--- a/cmd/link_package.go
+++ b/cmd/link_package.go
@@ -21,7 +21,7 @@ func LinkPackage(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interr
 	here := "../.." + strings.Repeat("/..", depth)
 
 	gopath := Gopath()
-	if len(gopath) == 0 {
+	if gopath == "" {
 		return nil, fmt.Errorf("$GOPATH appears to be unset")
 	}
 	if len(pname) == 0 {
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -115,11 +115,10 @@ func Gopath() string {
 // listed in the Gopath.
 func Gopaths() []string {
 	p := os.Getenv("GOPATH")
+	p = strings.Trim(p, string(filepath.ListSeparator))
 	ps := filepath.SplitList(p)
-
-	// XXX: Is this right? What is an empty path supposed to mean?
-	if ps[0] == "" {
-		ps[0] = "."
+	if len(ps) == 0 {
+		return []string{"."}
 	}
 	return ps
 }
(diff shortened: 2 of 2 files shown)


gcmt:
A bit of refactoring the first commit. Essentially I made both functions `Gopath()` and `Gopaths()` behave in a more predictable way.

cmd/link_package.go
```diff
@@ -21,7 +21,7 @@ func LinkPackage(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interr
 	here := "../.." + strings.Repeat("/..", depth)
 
 	gopath := Gopath()
-	if len(gopath) == 0 {
+	if gopath == "" {
```
technosophos: I'm curious as to why you changed the `len(gopath)` to a string comparison. Since even an empty string allocates a slice, I would think that the `len()` comparison would be faster. Of course, the compiler probably optimizes away any difference anyway.

I certainly wouldn't disapprove of the patch because of this, but I am curious why you'd make this change.

technosophos:
I'm good with this change. I think we can merge it.

technosophos:
lol... I just realized it was already merged. Thank you @gcmt !

cmd/link_package.go
gcmt: I just think that comparing `gopath` against and empty string makes it clear that we are expecting a string from `Gopath()` instead of a more general slice.

technosophos: Makes sense. I have an unhealthy interest in microoptimizations, so I was
wondering if it was a performance thing.

On Friday, September 25, 2015, Giacomo Comitti <notifications@github.com>
wrote:

> In cmd/link_package.go
> <https://github.com/Masterminds/glide/pull/89#discussion_r40433214>:
>
> > @@ -21,7 +21,7 @@ func LinkPackage(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interr
> >  	here := "../.." + strings.Repeat("/..", depth)
> >
> >  	gopath := Gopath()
> > -	if len(gopath) == 0 {
> > +	if gopath == "" {
>
> I just think that comparing gopath against and empty string makes it
> clear that we are expecting a string from Gopath() instead of a more
> general slice.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Masterminds/glide/pull/89/files#r40433214>.
>


-- 
http://technosophos.com
https://github.com/Masterminds

PR #98, opened 2015-10-09
Fix `import godep` subcommand reference

jonboulle:
The README lists the command as `import godep`, but that was resulting
in:

	Oops! Command glide-godep does not exist.

I think this is the right solution, but don't quite grok the cookoo
rabbit hole yet, so might be off base - there are a lot of godep/godeps
references around. This makes it work as promised at least :-).


technosophos:
Yes, this looks right. @mattfarina any objection to changing the command name?

mattfarina:
@jonboulle thanks for catching that and the pull request.

PR #114, opened 2015-10-21
Feature/semver

mattfarina:
Adding semantic version support


technosophos:
Nice! Wow... that was a lot of work. Looks good to me.

PR #115, opened 2015-10-22
Makefile: fix usage of $(VERSION)

dolmen:
Use the make variable from make, not as a shell variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ VERSION := $(shell git describe --tags)
 DIST_DIRS := find * -type d -exec
 
 build:
-	go build -o glide -ldflags "-X main.version=${VERSION}" glide.go
+	go build -o glide -ldflags "-X main.version=$(VERSION)" glide.go
 
 install: build
 	install -d ${DESTDIR}/usr/local/bin/
@@ -27,7 +27,7 @@ bootstrap-dist:
 
 build-all:
 	gox -verbose \
-	-ldflags "-X main.version=${VERSION}" \
+	-ldflags "-X main.version=$(VERSION)" \
 	-os="linux darwin windows " \
 	-arch="amd64 386" \
 	-output="dist/{{.OS}}-{{.Arch}}/{{.Dir}}" .


Makefile
```diff
@@ -2,7 +2,7 @@ VERSION := $(shell git describe --tags)
 DIST_DIRS := find * -type d -exec
 
 build:
-	go build -o glide -ldflags "-X main.version=${VERSION}" glide.go
+	go build -o glide -ldflags "-X main.version=$(VERSION)" glide.go
```
mattfarina: I'm not sure I understand the change here. `$(VERSION)` tells the shell script to execute `VERSION` and return the result of that. `${VERSION}` returns the value of the variable. In this case they will have the same result. But, I'm not sure of the motivation for the change.

Can you explain? Is there something about shell scripting I'm not aware of (which could easily be the case because I rarely do it)?

As the one who runs the builds, the current implementation generates the correct version. Are you seeing something different?

dolmen: > $(VERSION) tells the shell script to execute VERSION and return the result of that.

No `$(VERSION)` is expanded my `make` before the shell even sees it. `${VERSION}` is also expanded by `make` (before being seen by the shell), but as it is the same syntax as the shell, this is confusing (readers may expect like you did that the expansion is done by the shell) and must be avoided. To really let the shell see the reference to a `VERSION` environment variable, you have to write `$${VERSION}`, but this is not portable with non-Unix shells (for example, on Windows) while the expansion by `make` is.

You should read a [`make` manual](https://gnu.org/software/make/manual/html_node/Reference.html#Reference).

dolmen: > As the one who runs the builds, the current implementation generates the correct version. Are you seeing something different?

This is more a matter of style and portability to non-POSIX shell than an error.

mattfarina: What non-POSIX shell environment are you looking for support in?

dolmen: > What non-POSIX shell environment are you looking for support in?

Windows.

dolmen:
ping

technosophos:
I don't have any strong preference, but I don't like making changes with no actual reason other than stylistic preference.

Essentially there are two acceptable variants of the same thing. Portability is cited as a reason, but I don't see what makes `$()` more portable than `${}` when both are interpreted by `make`. What is the specific reason for merging this vs. keeping the current?

mattfarina:
@dolmen I'm not sure what the windows compatibility issue is. If you're using cygwin it shouldn't be an issue. `${}` is the syntax or variables in a Makefile. I've tested this on windows and have not found an issue.

Can you explain the reason for the switch in more detail. What am I not seeing?

PR #124, opened 2015-10-29
Feature/cache

mattfarina:
This uses either the GOPATH or the $HOME/.glide/cache to store repos. They are copied to the right location from there. This works with repos (as opposed to just package names so forks still work). The GOPATH is used only when the repo exactly matches.

By default the GOPATH is used as the cache unless there is something blocking or the `--cache` flag is used on `glide up` or `glide get`. This is to have one directory with all the caches for things like CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ distributed with the package.
     - hg
     - svn
 * Support custom local and global plugins (see docs/plugins.md)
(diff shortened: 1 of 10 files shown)


technosophos:
I read through this and it all looks like what I was envisioning when you described it to me. I think we should definitely merge this into 0.7.

PR #126, opened 2015-10-29
Add a --no-color flag to remove color codes from output

albrow:
This is incredibly useful for certain environments which do not support ansi color codes. For example, in my use case, we were running glide install in jenkins, where the color codes do not work and just add junk to the output.


mattfarina:
I've run into this exact problem with here and with other tools. Thanks for the contribution.

Issue #130, opened 2015-11-02
Support yml extension


h2non:
Both `yml` and `yaml` extensions are valid for YAML files. This change should be trivial, but helpful.

technosophos:
I'm not sure I understand the request. Where in particular where you looking for us to support something different?

If I recall correctly, the `-y` flag will take _any_ extension and still attempt to parse it as YAML.

The `-f` flag lets you set whatever filename you want, and will write to it regardless of the extension.

The default filename is `Glide.yaml`. I'm not sure it's a good idea to have more than one default filename, which is essentially what we'd get if we allowed `Glide.yaml` and `Glide.yml`. That would lead to unpredictable results if both a `Glide.yaml` and a `Glide.yml` file were present. But it's certainly possible today to do `glide -y Glide.yml up`.

(At one point, were were talking about naming the default file `Glidefile`. We added `.yaml` simply so that editors could provide syntax and highlighting.)

Thanks!

h2non:
IMHO this should be transparently supporting during the glide YAML file discovery process.
If  `glide.yaml` doesn't exists, fallback looking for `glide.yml`. Is that simple. Specific flag is great but useless except you're using a file with another completely different name.

mattfarina:
FYI, I'm looking at how we handle config files. Looking at a couple ways to change things up to make it easier.

dolmen:
@h2non I don't think that multiple names would help. So far you have not told us your reason why you want this feature.

If you want to call your file `Glide.yml` (maybe because your text editor prefers it), either fix your text editor or just make a symbolic link: `ln -s Glide.yaml Glide.yml`.

h2non:
Things are simpler. Officially both `.yml` and `.yaml` file extensions are registered and supported for the YAML language. 

If you need an real example of an extended use case of the simplified file extension I'm talking about you can take a look to this repository and verify that Travis config file is called `.travis.yml`.

We're just talking here about a trivial change that helps users significantly and provides a convenient implementation based on YAML community standards.
 
Resolving one or other file is trivial, and I can only list benefits, such as: doesn't introduces any kind of constraint in terms of the utility behavior, makes the utility more useful and robust for end users and it's less error prone.

mattfarina:
@h2non There are a couple things going on here. First, with the yaml file...

1. Isn't it that case that a majority of the time Glide is creating it for you?
2. What happens when you have a `glide.yaml` and `glide.yml` file?

[...]

PR #138, opened 2015-11-13
feature #130: support multiple manifest file extensions 

h2non:
Tests are basic. Tries to follow your existent code and interface conventions.

Needs a review and probably some extra tests. Take a look and let me know.

### DIFF
--- a/cmd/yaml.go
+++ b/cmd/yaml.go
@@ -13,7 +13,7 @@ import (
 )
 
 // YAML language compatible file extensions
-var YAMLExtensions = []string{".yml", ".yaml"}
+var YAMLExtensions = []string{".yaml", ".yml"}
 
 // ParseYaml parses the glide.yaml format and returns a Configuration object.
 //
@@ -42,7 +42,7 @@ func ParseYaml(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrup
 // in the current working directory
 func LookupManifestFile(fname string, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
 	ext := filepath.Ext(fname)
-	if ext != ".yml" && ext != ".yaml" {
+	if ext != YAMLExtensions[0] && ext != YAMLExtensions[1] {
 		return FileExists(fname)
 	}
 
--- a/cmd/yaml.go
+++ b/cmd/yaml.go
@@ -5,11 +5,15 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/Masterminds/cookoo"
 	"github.com/Masterminds/glide/yaml"
 )
+
+// YAML language compatible file extensions
+var YAMLExtensions = []string{".yml", ".yaml"}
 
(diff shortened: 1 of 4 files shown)


cmd/yaml.go
```diff
@@ -33,6 +38,46 @@ func ParseYaml(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrup
 	return cfg, nil
 }
 
+// LookupManifestFile look up for a given file
+// in the current working directory
+func LookupManifestFile(fname string, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
```
technosophos: So the way I read this, if I do `-y foo.yml` and I have two files, one named `foo.yml` and one named `foo.yaml`, then `foo.yaml` will get loaded even though I explicitly said `foo.yml`, right?

glide.go
```diff
@@ -319,8 +319,8 @@ Example:
 		{
 			Name:      "update",
 			ShortName: "up",
-			Aliases:   []string{"install"},
-			Usage:     "Update or install a project's dependencies",
+			//Aliases:   []string{"install"},
```
technosophos: Please leave the alias. We have it here for backward compatibility.

technosophos:
To speed up dev on this, I did a quick ack of the codebase for references to `glide.yaml`.

```
$ ack 'glide\.yaml' ./cmd/*.go
cmd/flatten.go
176:	gp := path.Join(dir, "glide.yaml")
193:	Info("Found glide.yaml in %s", gp)

cmd/get_imports.go
67:			Warn("Package %q is already in glide.yaml. Skipping", root)

cmd/init_glide.go
33:// Among other things, it creates a default glide.yaml.
36:// 	- filename (string): The name of the glide YAML file. Default is glide.yaml.
39:	fname := p.Get("filename", "glide.yaml").(string)

cmd/link_package.go
29:		return nil, fmt.Errorf("glide.yaml is missing 'package:'")

cmd/util.go
37:// Most importantly, it fails if glide.yaml is not present in the current
40:	fname := p.Get("filename", "glide.yaml").(string)

cmd/vendor_path.go
14:	filename := c.Get("yaml", "glide.yaml").(string)
20:	// Find the directory that contains glide.yaml

cmd/vendor_path_test.go
11:	filename := "glide.yaml"

cmd/yaml.go
14:// ParseYaml parses the glide.yaml format and returns a Configuration object.
22:	fname := p.Get("filename", "glide.yaml").(string)
102:			Warn("Package %s is already in glide.yaml. Skipping.\n", dep.Name)
```

cmd/yaml.go
h2non: I've just pushed a new commit which should look first at `.yaml`, and then to `.yml`, so now you're right. Having multiple files with the same name a different extension is an edge scenario and a mistake by the end user.

glide.go
h2non: Done!

mattfarina:
This PR was incomplete and is no longer compatible with the current codebase so I'm going to close it. If someone wants to pursue this change we're happy to discuss it. Closing due to the state and it's comparison to master.

Issue #143, opened 2015-11-18
Planning for pinning


sdboyer:
To help ensure we do pinning properly, @mattfarina @technosophos and I spent a bit of time on videochat, and we drew this purty thing, and I'm putting it up here, along with my understanding of what we laid out, for reference:

![img_20151118_122939](https://cloud.githubusercontent.com/assets/21599/11249757/d0db881e-8df6-11e5-9db7-af669f88e158.jpg)

The matrix describes the set of possible disk states glide might encounter, against the set of possible commands a user might run. Our goal was to articulate a basic class of behavior for each of those combinations. 

For this diagram to make sense, it's important to understand the overall strategy glide is pursuing: *glide ALWAYS tries to create the most reproducible build possible.* This entails that, for each `main` package and its corresponding `glide.yaml`, all dependencies (direct and transitive) should be pinned to specific, immutable versions. There's no wild-westing, managing some external packages but not others; glide assumes and enforces that it 

This fixed objective allows glide to become much *simpler* and more predictable in its operations. It allows us say that, for every command that is run, we target exactly one class of output disk states (in the matrix, the state with an orange star):
* a valid `glide.yaml` file
* lock file that is correct with respect to the version bounds established in the `glide.yaml` file
* vendor directory containing packages that match exactly the pinned versions described in the lock file

Because there is only the one class of target output states, it removes ambiguity about how glide should deal with different types of user actions. This is not only because the end goal is clear, but also because it is impossible that glide would have itself intentionally left the disk in anything other than the correct output state - and therefore, any on-disk state glide finds represents user intent.

These rules amount to removing circular dependencies. M. C. Escher is great, but you don't want him in your package manager.

The disk states in the matrix aren't quite exhaustive, but it's pretty good. They are:
* **No glide.yaml**: basically, when you first start out. No glide.yaml implies no lock file and no vendor (if they exist, they are considered detritus)
* **Has glide.yaml, no lock file**:
* **Has glide.yaml, outdated lock file**: the lock file always contains a hash of the dependencies from the version of the glide.yaml file from which it was generated. If the hash does not match, we know the lock file is out of date.
* **Has glide.yaml, lock file, wrong local state**: there is a discrepancy between what's in the lock file and what's checked out in the vendor directory.
* **Has glide.yaml, lock file, correct local state**: this is the 'blessed' state that all commands seek to reach.
* **Has empty glide.yaml and .go files contain some external dependencies**: commonly the case when starting using glide on a new project.
[...]

PR #144, opened 2015-11-20
Cleanup delete.go a bit

imikushin:
Add delete command and cleanup delete.go a bit

Delete dir if the package has subpackages and the dir is not in a declared subpackage. 
If a subpackage is (or ends with) '.', then it's subdirs won't be considered to be used.

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -30,7 +30,12 @@ func DeleteUnusedPackages(c cookoo.Context, p *cookoo.Params) (interface{}, cook
 	cfg := p.Get("conf", nil).(*yaml.Config)
 	var pkgList []string
 	for _, dep := range cfg.Imports {
-		pkgList = append(pkgList, dep.Name)
+		for _, sub := range dep.Subpackages {
+			pkgList = append(pkgList, dep.Name+"/"+sub)
+		}
+		if len(dep.Subpackages) == 0 {
+			pkgList = append(pkgList, dep.Name)
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


imikushin:
I needed to remove unused packages after `glide up`, because I vendor dependencies and want to have leaner checkouts, but didn't want to mess up with what people are used to.

Usage: `glide del`

mattfarina:
@imikushin thanks for the pull request.

We're working on a change for locking/pinning (#143 for details) and this includes prune functionality. I'd like to hold off a few weeks on this until we get far along in that so the prune feature can be tied to that. The branch that work is happening on is quickly diverging from master. We're trying to keep master stable because so many people pull from that directly rather than a release.

That ok for you?

imikushin:
@mattfarina I'm looking forward to #143 being implemented. 

I'll remove the `delete` command from this PR and will only leave the small fixes to `delete.go`. They are basically dealing with path prefixes:
e.g. `github.com/docker/docker` is a valid path prefix for `github.com/docker/docker/api`, but not for `github.com/docker/dockerinit`

I guess, I'll have to use my fork for a while to delete unused packages from vendor: we (@ Rancher Labs) have a policy to store the vendored libs under version control, so the prune feature is pretty much a requirement.

technosophos:
This looks pretty straightforward to me. I don't understand why we'd need to loop over subpackages separately, though.

mattfarina:
This PR is no longer compatible with the current state of the codebase. A PR against the current state of the codebase (there were major changes) is something we'd be happy to look at. Closing.

PR #148, opened 2015-11-25
Make glide "go installable"

paulmach:
This is a suggestion to make glide "go installable" by copying in the dependencies directly to vendor directory. I like the submodule vendoring used by glide, but in order to be "go installable" glide will need to copy in the files, the burden of bootstrapping I guess.  

For developers using glide the install will look something like:

* install go, with two extra things requires that most should already be doing
  * export GO15VENDOREXPERIMENT=1
  * export PATH=$PATH:$GOPATH/bin
* go install github.com/Masterminds/glide

For glide developers, they need to have the vendor experiment enabled. When deps need to be updated just run `make bootsrap` to recopy in the latest deps into the vendor directory.

The pros are, consistency with other tools like golint and the deps are actually locked, the current makefile does not guarantee this. 

I didn't see any references to this issue in open or closed issues and prs so do not know if it's been suggested before. I'm curious what you guys think about this. 

Still todo:
* [ ] Remove some other commands from the makefile
* [ ] Update the readme

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 _vendor
-vendor/
 glide
 *.a
 *.sublime-project
--- a/Makefile
+++ b/Makefile
@@ -17,12 +17,19 @@ clean:
 	rm -rf ./dist
 
 bootstrap:
+	rm -rf ./vendor
 	mkdir ./vendor
+	
 	git clone https://github.com/Masterminds/cookoo vendor/github.com/Masterminds/cookoo
+	rm -rf vendor/github.com/Masterminds/cookoo/.git
 	git clone https://github.com/Masterminds/vcs vendor/github.com/Masterminds/vcs
+	rm -rf vendor/github.com/Masterminds/vcs/.git
 	git clone https://gopkg.in/yaml.v2 vendor/gopkg.in/yaml.v2
+	rm -rf vendor/gopkg.in/yaml.v2/.git
 	git clone https://github.com/codegangsta/cli vendor/github.com/codegangsta/cli
+	rm -rf vendor/github.com/codegangsta/cli/.git
 	git clone https://github.com/Masterminds/semver vendor/github.com/Masterminds/semver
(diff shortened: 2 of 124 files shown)


mattfarina:
@paulmach thanks for bringing this up.

There are some fairly large cons that should be noted.

* The codebase is no longer just the application code. It's the application plus a copy of all dependencies.
* The repo will grow significantly larger.

You may be able to see my preference to not store outside dependencies in another projects repo.

You are right that the current bootstrap process isn't ideal because of versions. When 0.8 comes out we'll have something in better shape.

While we have a `glide.yaml` file for glide itself we're trying to keep the tip of master installable. Is there something not working with `go install github.com/Masterminds/glide` today?

paulmach:
:100:% agree with the cons, that's why we make so much use of glide. I guess I'm saying that the pros outweigh the cons for a core utility like glide. Having go installablity with the proper dependencies and thus embracing the go tool chain is a major pro.

`go install` works, but you're not locking to specific dependency versions. Which is kind of weird for a dependency manager. But I guess that could be fixed by add some `git reset --hard sha` to the makefile, but then there is still a makefile.

mattfarina:
I'll talk with @technosophos about it. This is something we'll need to carefully consider.

technosophos:
I'm still interested in solving this problem. @mattfarina and I have been having some conversations about changes we need to make to Glide in order to systematically support this kind of requirement. Having Glide dogfood the approach is desirable.

mattfarina:
@technosophos now that the `--strip-vcs` flag is in Glide should we start to vendor as outlined here?

technosophos:
I think I'd be okay with testing this with `--strip-vcs` and a committed `vendor/` directory. With Go 1.6, that should work just fine.

mattfarina:
This PR has become out of date and can't be merged. I've setup the same functionality at #352. Due to the stale nature of this request I'm going to close.

Issue #153, opened 2015-12-06
Readme fie should change  with the new changes of Glide


kgrvamsi:
With the newer version of glide which supports Go 1.5 and most of the options which are been depreciated have to be updated on the readme file.

mattfarina:
We've made a couple passes on the readme with updates.

Issue #99, opened 2015-10-09
`glide get` should pin version by default


albrow:
Currently, when you run `glide get`, it adds a line to glide.yaml which looks something like:

```
import:
  - package: github.com/Masterminds/cookoo
```

`glide get` should pin the version by default by including the `ref` in glide.yaml. I cannot imagine a use case where you would want to add a package to glide.yaml and *not* pin the version.

In one of my projects we are using glide and not committing the vendor directory to git. If there is no version for a package in glide.yaml, it means our builds are not reproducible. So I have had to go in to glide.yaml and manually add the version each time.

I understand that this is not the same workflow that everyone uses. If you are committing the vendor directory to version control, it certainly won't harm you to have the version pinned by default in glide.yaml.

mattfarina:
Thanks for the feedback here.

My original thought was to specify a version as a flag. For example, `glide get github.com/Masterminds/cookoo --version 1.2.0`. But, over on #96 there was a discussion of retrieving more than one dependency with `glide get`. That changes how you can handle dealing with versions.

You are very right that not specifying a version leads to unreproducible builds. This is an area we'd like to improve the experience.

albrow:
@mattfarina thanks for the quick response. I like the idea of specifying a version (or commit hash for projects not using SemVer) with the `--version` flag.

This issue still has relevance for cases where the `--version` flag is not provided. In that case, glide should assume you want the latest version and should add version information to glide.yaml to prevent the version from changing between builds. If the project is not using SemVer, `glide get` should pin `ref` to the latest commit on the master branch. If the project is using SemVer, `glide get` should add the latest major or minor version to glide.yaml. This is how bundler behaves in the ruby world and how npm behaves for javascript (except they always assume SemVer because of the way packages are added to the registry).

technosophos:
The main design of `glide get` was to mirror `go get` quite closely, but put the downloaded packages into `vendor`, and track it in `glide.yaml`. Pinning is an additional step that isn't really close to the default Go way of doing things (not that that's right -- just that it's expected).

That said, I can understand why many developers prefer to pin their dependencies. Some even want to always keep them pinned. To facilitate that, I'd be okay with a `--version` or `--pin, -p` flag. I would not be comfortable with moving the default `glide get` behavior further from `go get`, though.

I think @albrow might have given us a suitable workaround for our previous `--version` problem. If `--pin` is specified, we could simply pin _all_ of the specified dependencies.

So assume this:
```
[...]

Issue #85, opened 2015-09-20
Support a style of npm shrinkwrap


itsjamie:
`glide pin` would change to directly generating a `glide-pin.yaml` file.
`glide install` would check if a `glide-pin.yaml` file existed, and if it did would install using that instead of `glide.yaml`
`glide up(date)` would reference `glide.yaml` and would update the underlying repositories, hopefully having the semantic version support discussed in #52.

This can be achieved currently using the global -y option, but I think having in-tool support would make adoption easier. Especially when recursing through other packages that might have glide inside them, there is no option for a user to choose which YAML file to use, so in the case of someone rolling this themselves, they would require their default yaml file be the pinned version.

Thanks for any consideration.

technosophos:
I've been playing with Elixir this weekend, and I like the fact that their tool has support for dev, test, and prod "environments".

I wonder if we could borrow a fragment of that, yet basically keep the same idea as is presented here.

For example, we could say that by default, Glide is in `dev` mode, but provide a global flag for setting modes. `glide -m prod` would explicitly be prod mode, while `glide -m dev` would explicitly be dev mode.

Each mode would by default first look for `glide-MODE.yaml` (or some similar pattern TBD), and then fall back to `glide.yaml`.

Then `glide pin -m prod` would generate `glide-prod.yaml`. While `glide up -m prod` would update based on the `glide-prod.yaml` file (if it exists).

If we were to go this way, what really are the modes? Ruby and Elixir break down into prod, dev, and test. Those seem reasonable to me... though I'm not really sure what the use case for `test` would be.

itsjamie:
Rather than allowing any mode, I would limit it to either "pinned" or not.

I think this is a case where less is actually more. If I could think of another use-case where I would want to have multiple pinned dependencies for my codebase based on the environment then I would get this.. but whatever your environments look like for your product, you should really only ever have two states ongoing at any point in your codebase.

Something bound for prod, which might hit many environments on it's way to prod. Such as... `integration -> qa -> staging -> prod` and then the current dev build. Anything bound for prod in our use-case would be pinned the whole way, and the dependencies version shouldn't change based on what environment they are in (though their configuration might). 

Does that make sense?

mattfarina:
@itsjamie thanks for bringing this up.

While I have my opinion on workflow and i'm sure @technosophos does as well, I'm not sure Glide should enforce just one. Instead I'd like it to enable developers. Sometimes there cases, even edge cases, that you may need something like this for.

[...]

PR #156, opened 2015-12-07
Feat/lockfile

mattfarina:
Add the ability to have a `glide.lock` file pinning to specific versions and install from that. This re-introduces the `glide install` command that install from a lockfile while `glide update` updates that material.

This touches a bunch of open issues including #153, #143, #99, and #85.


technosophos:
I've done a pretty thorough test of this on pipeviz, Deis Helm, Deis Etcd, and Glide itself.

Issue #182, opened 2015-12-21
glide handling of packages with the same prefix


dmitris:
I wonder if glide has any special logic to consider imports like "package-with-dashes-one" and "package-with-dashes-two" to be really subpackages of "package-with-dashes" for any reason?

I have a test project sometld.com/dmitris/go-simple-canary-cmd
that depends on sometld.com/dmitris/go-simple-canary-lib and through that library transitively on
sometld.com/dmitris/go-simple-canary-lib-two and sometld.com/dmitris/go-simple-canary-lib-three (for testing the CI system).

When I run "glide init" and "glide up", it creates a glide.lock file with:
```
imports:
- name: sometld.com/dmitris/go-simple-canary-lib
  version: 50cfa0f24c75a099648c023a5548ea37f20e9fdc
  subpackages:
  - -two
  - -three
```

glide.yaml has only "sometld.com/dmitris/go-simple-canary-lib" even when I explicitly try to run "glide get sometld.com/dmitris/go-simple-canary-lib-two".

Initially, I had packages named like go-simple-canary-lib, go-simple-canary-lib2, and go-simple-canary-lib3 and glide did a similar thing: it put "2" and "3" as subpackages even though each one a stand-alone package with its own import path...

mattfarina:
Glide has special logic to find the top of VCS repos. Are all of these in the same VCS repo?

One place this happens inside [GetRootFromPackage](https://github.com/Masterminds/glide/blob/8b683f9ec07fb4463a252ead1ee91b693f1c4cb9/util/util.go#L30). In here it tries a variety of methods to figure out the root repo.

The sub-packages are the suffix to the root package. You can see some of this logic at work in [NormalizeName](https://github.com/Masterminds/glide/blob/8b683f9ec07fb4463a252ead1ee91b693f1c4cb9/cmd/yaml.go#L163).

Can you share more detail about your layout? What VCS? How is it laid out? Is this private or public? Do you have a go-get redirect in place?

dmitris:
Yes, they are all the same VCS (git).

The layout is:
go-simple-canary-cmd
go-simple-canary-lib
go-simple-canary-lib-two
go-simple-canary-lib-three

each one is part of the package name like "go.corp.company.com/dmitris/go-simple-canary-cmd" (the domain is not the real one).  The dependency chain is: cmd -> lib -> lib-two -> lib-three (toy projects to test dependency resolution by the CI system).

'go list -e -f' finds them all fine:
```
[...]

PR #183, opened 2015-12-22
check for exact match instead of string prefix for #182

dmitris:
this should avoid truncating packages in cases when one package is a substring of another one (without being a "subpackage") - for example, "foobarxyz.com/user/lib" and "foobarxyz.com/user/lib2"


mattfarina:
@dmitris thanks for finding this issue.

For anyone else looking for the commentary see #182.

Issue #166, opened 2015-12-14
Vendor the dependencies of the sub-packages only, not of the parent package


kshlm:
The dependent packages can have subpackages, but what is their use? They don't seem to be used any where.

For example, my project uses `github.com/coreos/etcd/client`. I have the following entry as a dependency,
```yaml
- package: github.com/coreos/etcd
  version: v2.2.2
  subpackages:
  - /client
```

I had assumed that because subpackages had been provided, glide would just resolve the dependencies of the subpackages. But instead, glide resolves dependencies of etcd and pulls in way too many dependent packages, which I have no use of in my project.

I want to know if this is the expected behaviour?

kshlm:
Should I rather just use the following,

```yaml
- package: github.com/coreos/etcd/client
  version: v2.2.2
```

GeertJohan:
Glide /could/ use the set of subpackages to only recursively vendor dependencies of the mentioned subpackages, instead of recursively vendoring the imports for all subfolders. See #165 for reasons why.

mattfarina:
Working with Go packages and versions is a bit complicated because there is technically no higher level "project" grouping. Version control systems group packages at a higher level and versions are at this level. So, you can't (and shouldn't) have two different versions of packages within `github.com/coreos/etcd`.

When you use `go get` or the other tools that fetch dependencies they do so at the repo level. They have to get everything. Try cloning a subdirectory from a git repo on GitHub. It doesn't work that way. So, have to fetch everything from a repo.

We could be better with sub-package handling for cleanup and walking the tree. Our first priority is to cleanup some internals in Glide so it's easier to maintain and contribute to. Better handling will come.

Part of the reason we have issues like this is the use of monorepos where clients and everything is in one repo. It may be easier for the developers (or they think it is) of the project. But, for consumers of things like a client it's harder. I had not run into such a prevalence of clients being part of the repo/project it communicates with until working in Go.

kshlm:
@mattfarina I just wanted to confirm what the expected behaviour of subpackages is.

I think we agree that using subpackages should cause only the dependencies of the subpackages should be vendored.

This is not how glide works currently, but is something that is being aimed at in a later release. Can we convert this issue to an enhancement request then?

GeertJohan:
> When you use `go get` or the other tools that fetch dependencies they do so at the repo level. They have to get everything. Try cloning a subdirectory from a git repo on GitHub. It doesn't work that way. So, have to fetch everything from a repo.

[...]

Issue #165, opened 2015-12-14
glide panics trying to generate lockfile with `--no-recursive`


kshlm:
```shell
$ glide --version
glide version 0.8.0
```

I'm using glide to manage dependencies for a project of mine. The recursive update mechanism enabled in 0.8.0, causes glide to download a huge number of unneeded packages. I would like to add only the required packages to glide.yaml and use `--no-recursive` to download just them. But using `--no-recursive` causes update to fail with a panic, `panic: interface conversion: interface {} is nil, not *cfg.Lockfile` (the full backtrace is at the end).

I use `github.com/docker/libkv` which has a sort of plugin architecture, which allows me to choose from multiple backends (etcd, consul, bolt and zookeeper). As I intend to use just one of the backends, the others don't matter to me. Glide pulls in all the dependencies for libkv, including etcd, consul and their large list of dependencies. I want to use `--no-recursive` to avoid this.

```
panic: interface conversion: interface {} is nil, not *cfg.Lockfile

goroutine 1 [running]:
github.com/Masterminds/glide/cmd.WriteLock(0x7f25c17a86c8, 0xc820111410, 0xc820032318, 0x0, 0x0, 0x0, 0x0)
        /Users/mfarina/Code/go/src/github.com/Masterminds/glide/cmd/yaml.go:108 +0x456
github.com/Masterminds/glide/vendor/github.com/Masterminds/cookoo.(*Router).doCommand(0xc82013cea0, 0xc820160810, 0x7f25c17a86c8, 0xc820111410, 0x0, 0x0, 0x0, 0x0)
        /Users/mfarina/Code/go/src/github.com/Masterminds/glide/vendor/github.com/Masterminds/cookoo/router.go:209 +0x86
github.com/Masterminds/glide/vendor/github.com/Masterminds/cookoo.(*Router).runRoute(0xc82013cea0, 0x9508b8, 0x6, 0x7f25c17a86c8, 0xc820111410, 0x950800, 0x0, 0x0)
        /Users/mfarina/Code/go/src/github.com/Masterminds/glide/vendor/github.com/Masterminds/cookoo/router.go:164 +0x4c9
github.com/Masterminds/glide/vendor/github.com/Masterminds/cookoo.(*Router).HandleRequest(0xc82013cea0, 0x9508b8, 0x6, 0x7f25c17a86c8, 0xc820111410, 0xc8200eb000, 0x0, 0x0)
[...]

PR #196, opened 2015-12-29
Generate lockfile when not vendoring recursively

kshlm:
The pin command has been removed, so lockfiles are useful for pinning
packages even if we are not vendoring recursively. This helps using a
simple `glide get` command as an alternative to doing a `glide get`
followed by manual pinning in glide.yaml.

### DIFF
--- a/cmd/flatten.go
+++ b/cmd/flatten.go
@@ -33,10 +33,6 @@ func Flatten(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt)
 	cacheGopath := p.Get("cacheGopath", false).(bool)
 	useGopath := p.Get("useGopath", false).(bool)
 
-	if skip {
-		Warn("Skipping lockfile generation because full dependency tree is not being calculated")
-		return conf, nil
-	}
 	packages := p.Get("packages", []string{}).([]string)
 
 	// Operate on a clone of the conf so any changes don't impact later operations.
@@ -71,6 +67,13 @@ func Flatten(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt)
 	deps := make(map[string]*cfg.Dependency, len(confcopy.Imports))
 	for _, imp := range confcopy.Imports {
 		deps[imp.Name] = imp
+	}
+
+	if skip {
+		Warn("Skipping recursive flattening")
+		Info("Project relies on %d dependencies.", len(deps))
+		c.Put("Lockfile", cfg.LockfileFromMap(deps, hash))
+		return conf, nil
 	}
 
 	f := &flattening{confcopy, vend, vend, deps, packages}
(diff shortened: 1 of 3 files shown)


kshlm:
Lockfile generation was disable for non-recursive vendoring as a fix for #165. But lockfile are useful for pinning packages as the `glide pin` command has been removed now.

@mattfarina What do you think?

mattfarina:
@kshlm Can you help me understand your use case?

I see two cases. One where you need to snapshot and restore the entire dependency tree. This is important for reproducible builds. To have a reproducible build you have to have the complete dependency tree. The complete tree in the `glide.lock` file is used by `glide install`. If the complete tree isn't available you can't restore to a known state. State is important and without the whole tree you don't have a complete record of the state to reproducibly restore to.

The other is the case to have info your project. In that case why would pinning to commit id be useful?

kshlm:
As I'd mentioned in #165, I do not want to pull in the huge dependencies for things I don't use. The only way to do it till #166 is done is to use `--no-recursive` and manually add the required dependencies to `glide.yaml`. Without the lock file I would need  to manually pin in `glide.yaml`, so the `glide up` installs the correct versions. I'm okay with manually pinning packages with versions in `glide.yaml`, but do not want to add commit-hashes manually. `glide pin` helped with this earlier. Now I have no way.

kshlm:
My main reason for getting this to work is to avoid downloading ~100 dependencies several of which are large, when I know my build works just fine with ~11. Having the smaller set of dependencies is a huge speed up for me.

mattfarina:
@kshlm thanks for the details

@technosophos what do you think about this until we can automatically detect the code paths being used?

kshlm:
@technosophos Could you take a look please? Thank you.

mattfarina:
@kshlm I've not forgotten about this. @technosophos has been pulled into some meetings and has not been available this week. Hope to get some of his time next week to discuss this.

emilevauge:
This would be very useful because recursive vendoring could be unusable in some cases: in http://github.com/EmileVauge/traefik, glide is downloading more than 600 Mo in `vendor/` :'(

mattfarina:
@emilevauge We (@technosophos actually) is working to make dependency smarter to look at the ones actually being used. That should lighten the load while still fetching the in use tree.

technosophos:
@kshlm I am so sorry for letting this PR languish for so long. Is this something you still need after that massive change I did with 0.9?

Again, I am so sorry for leaving this out here.

kshlm:
@technosophos With glide-0.9, I don't require this. But I still think, having lockfiles for non-recursive vendoring is useful, when trying to get reproducible builds.

mattfarina:
@kshlm Can you describe a base where you would want a reproducible build and not wanting to recurse? I'm having a little trouble visualizing this one. An example would help me.

kshlm:
Closing this old PR. This is not applicable anymore, and it is not really needed anymore.

PR #200, opened 2015-12-29
Fix example glide.yaml in README.

jrick:
subpackages directive must be plural.

Contents of the subpackages directive must be a yaml list.


mattfarina:
Thanks for catching this documentation bug.

PR #220, opened 2016-01-21
Vendor only the subpackage(s) dependencies

hectorj:
Implements #166 

WIP, not for merging as is.

Using @davecgh 's [test](https://github.com/Masterminds/glide/issues/166#issuecomment-168332303), with the current code in this PR I get:
```
$ tree -d vendor
vendor
└── github.com
    └── syndtr
        └── goleveldb
            ├── leveldb
            │   ├── cache
            │   ├── comparer
            │   ├── errors
            │   ├── filter
            │   ├── iterator
            │   ├── journal
            │   ├── memdb
            │   ├── opt
            │   ├── storage
            │   ├── table
            │   ├── testutil
            │   └── util
            └── manualtest
                ├── dbstress
                └── filelock

19 directories
```

### DIFF
--- a/dependency/resolver.go
+++ b/dependency/resolver.go
@@ -283,15 +283,35 @@ func (r *Resolver) ResolveAll(deps []*cfg.Dependency) ([]string, error) {
 // resolveList takes a list and resolves it.
 func (r *Resolver) resolveList(queue *list.List) ([]string, error) {
 
+	res := make([]string, 0, queue.Len())
+
 	var failedDep string
 	for e := queue.Front(); e != nil; e = e.Next() {
 		dep := e.Value.(string)
-		t := strings.TrimPrefix(e.Value.(string), r.VendorDir+string(os.PathSeparator))
-		if r.Config.HasIgnore(t) {
-			msg.Info("Ignoring: %s", t)
+		importPath := r.vendorToImportPath(dep)
+
+		// If subpackages are specified, process only them, not the root
+		existing := r.Config.Imports.Get(importPath)
+		if existing != nil && len(existing.Subpackages) > 0 {
+			// Build the list of subpackages
+			sublist := list.New()
+			for _, subPkg := range existing.Subpackages {
+				sublist.PushBack(filepath.Join(dep, filepath.FromSlash(subPkg)))
+			}
+			// Process it
+			subRes, err := r.resolveList(sublist)
+			if err != nil {
+				return []string{}, err
+			}
+			res = append(res, subRes...)
 			continue
 		}
-		r.VersionHandler.Process(t)
+
+		if r.Config.HasIgnore(importPath) {
+			msg.Info("Ignoring: %s", importPath)
+			continue
+		}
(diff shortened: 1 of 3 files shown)


hectorj:
https://github.com/Masterminds/glide/issues/166#issuecomment-164742093

> It would change the way scanning works if we encounter projects from Godeps, GPM, and GB which glide can import from.

The choice made here is to only apply this new logic only if no Godeps, GPM nor GB file is found (so only for `mergeGuess`)

I tested it on one of our internal project, and it seems to work well, reducing the number of dependencies by ~ 1/3 in our case which is nice.

mattfarina:
Sometime soon, likely next Monday so I can be around to fix bugs as they arise, we're going to merge in the `feat/no-cookoo` branch. This is a massive internal reworking of Glide to make it easier to maintain and work on. There was a lot of left over cruft in the codebase from the past.

Unfortunately, that will completely invalidate this pull request. The `cmd` directory won't even exist anymore.

hectorj:
Noted, I 'll try to redo my work on top of that branch, thanks.

hectorj:
Ok, I managed to do the same thing on the new master branch.

Still gotta add the flag to make it optional

hectorj:
Ok, can you do a first review and tell me what you think of it?

dependency/resolver.go
```diff
@@ -283,15 +284,37 @@ func (r *Resolver) ResolveAll(deps []*cfg.Dependency) ([]string, error) {
...
+				}
+				// Process it
+				subRes, err := r.resolveList(sublist)
+				if err != nil {
+					return []string{}, err
+				}
+				res = append(res, subRes...)
+				continue
+			}
+		}
```
mattfarina: This is looking at the sub-packages in the config while still walking the directory tree (`filepath.Walk` below).

Instead of walking the file tree for packages in the `vendor/` folder what about walking the Import tree? This is what @technosophos is working on at the moment.

I'd like to hold off on this until we get to see what that looks like so we can compare them.

fabian-z:
This branch is working fine for me when merged on top of master. I searched for a solution because glide suddenly imported 44 dependencies for my project. Using this feature flag cuts it down on 20 with identical functionality (and I believe identical with go get, but I will have to check).

mattfarina:
@fabian-z The implementation details can matter. For example, the amount of work being done overall inside the applications can be very different to come to the same outcome. Want to make sure we use the best approach if there are multiple being developed.

mattfarina:
@hectorj @fabian-z On the pull request at #240 we have a walker capable of just walking the imports rather than the whole file tree. One advantage is that it only walks what's used rather than walking everything and filtering down to what's used.

Any thoughts?

hectorj:
Haven't tested, yet but that sounds great and way better than this PR

mattfarina:
I just merged the other branch. Thanks for the efforts here. That method is a little more code but it is less of the tree to walk. We can build upon and improve things over there.

Issue #175, opened 2015-12-18
win glide 0.8.1 doesn't build github.com/aws/aws-sdk-go


payneio:
win glide 0.7.2 works

0.8.1 gets stuck on checking dependencies

mattfarina:
@payneio thanks for the info. Can you tell me what version of Windows you're on, what your setup looks like (is the parent project in the `GOPATH`, etc), how you got Glide (was it a binary download or `go get`) and what architecture (386/amd64)? Does it work for projects that aren't the aws-sdk-go?

I'll try to reproduce the problem.

flawedmatrix:
I have the same problem on OS X (10.10.4).

```bash
$ go version
go version go1.5.2 darwin/amd64
$ brew install glide
...
$ glide --version
glide version 0.8.1
$ mkdir test_workspace
$ cd test_workspace/
$ export GOPATH=$PWD
$ export PATH=$GOPATH/bin:$PATH
$ export GO15VENDOREXPERIMENT=1
$ glide create
[INFO] Generating a YAML configuration file and guessing the dependencies
[INFO] Attempting to import from other package managers (use --skip-import to skip)
$ glide get github.com/aws/aws-sdk-go
[INFO] Preparing to install 1 package.
[INFO] Importing github.com/aws/aws-sdk-go
[INFO] Walking the dependency tree to calculate versions
[INFO] Fetching updates for github.com/aws/aws-sdk-go.
[INFO] Scanning github.com/aws/aws-sdk-go for dependencies.
[INFO] ==> Unknown github.com/aws/aws-sdk-go (github.com/aws/aws-sdk-go/aws/awserr)
[INFO] ==> Unknown github.com/aws/aws-sdk-go (github.com/aws/aws-sdk-go/aws/credentials)
[INFO] ==> Unknown github.com/aws/aws-sdk-go (github.com/aws/aws-sdk-go/aws)
[INFO] ==> Unknown github.com/aws/aws-sdk-go (github.com/aws/aws-sdk-go/aws/client/metadata)
[INFO] ==> Unknown github.com/aws/aws-sdk-go (github.com/aws/aws-sdk-go/aws/request)
[INFO] ==> Unknown github.com/aws/aws-sdk-go (github.com/aws/aws-sdk-go/aws)
[INFO] ==> Unknown github.com/aws/aws-sdk-go (github.com/aws/aws-sdk-go/aws/awserr)
[...]

PR #224, opened 2016-01-23
Fix issue #175 preventing lookup of packages directly inside the vendor di…

andrewzeneski:
…rectory

This addresses issues getting packages that fall under the category of "unknown" package type.


andrewzeneski:
This resolves #175 for me.

andrewzeneski:
Turns out I didn't fully understand where the "vendor" directory was supposed to go. This pull request turns out to be moot. Please disregard.

mattfarina:
No problem. This `vendor/` directory thing is going to take some getting used to.

PR #240, opened 2016-02-04
Resolve the imports being used

mattfarina:
This walks the import tree rather than the file system to find the imports to download.

Nice work @technosophos 

I think there are two things still to do:

- [x] Remove some of the message output that should be debug instead.
- [x] Glide has the ability to walk the file system and the import path. Choosing could be a command line switch (which could be useful in development).

### DIFF
--- a/dependency/resolver.go
+++ b/dependency/resolver.go
@@ -356,6 +356,7 @@ func (r *Resolver) resolveImports(queue *list.List) ([]string, error) {
 			if ok, err := r.Handler.NotFound(dep); ok {
 				r.alreadyQ[dep] = true
 				queue.PushBack(r.vpath(dep))
+				r.VersionHandler.SetVersion(dep)
 			} else if err != nil {
 				msg.Warn("Error looking for %s: %s", dep, err)
 			} else {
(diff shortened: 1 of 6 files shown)


technosophos:
Which seems like it should be the default? I'd go for the import path as the default, and the file walker as an option.

mattfarina:
@technosophos I was thinking the same thing for the switch. The switch is useful in some dev cases when you want all possible things.

dependency/resolver.go
```diff
@@ -277,7 +280,157 @@ func (r *Resolver) ResolveLocal(deep bool) ([]string, error) {
...
+				//// This was already processed, possibly by another
+				//// concurrent iteration.
+				//continue
+				//}
+				r.alreadyQ[dep] = true
+				queue.PushBack(r.vpath(dep))
+			} else if err != nil {
+				msg.Warn("Error looking for %s: %s", dep, err)
+			} else {
+				// TODO (mpb): Should we toss this into a Handler to
```
technosophos: @mattfarina What should we do here?

Issue #264, opened 2016-02-17
Caching should be disabled when flag is not present


mhoglan:
Currently it looks like caching is disabled if you do not pass in the `--cache` flag, but really it is just erroring out because the directory does not exist and skips the caching step.

This is because the `cacheEnabled` var in repo/cache.go is always set to true and is never changed to false based on the `Installer.UseCache` value.

I explicitly set it to false in the code and rebuilt to see the difference.   I am not sure where the best place to set the var to false is.  The `cache.go` is not an object you interact with, but utility functions, so there is no singleton to change the value.   Could have the `installer.go` set the var when it is called on since it is in the same scope of the package, but that does not satisfy other files in the package if the installer is never called.   

When watching the DEBUG printout you can see the difference

```
[DEBUG] Attempting to find current branch for https://github.com/jmespath/go-jmespath
[DEBUG] Saving default branch for https://github.com/jmespath/go-jmespath
[DEBUG] Error saving https://github.com/jmespath/go-jmespath to cache - Error: open /root/.glide/cache/info/https-github.com-jmespath-go-jmespath.json: no such file or directory
[DEBUG] Attempting to find current branch for https://github.com/spf13/pflag
[DEBUG] Saving default branch for https://github.com/spf13/pflag
[DEBUG] Error saving https://github.com/spf13/pflag to cache - Error: open /root/.glide/cache/info/https-github.com-spf13-pflag.json: no such file or directory
[DEBUG] Attempting to find current branch for https://github.com/spf13/viper
[DEBUG] Saving default branch for https://github.com/spf13/viper
[DEBUG] Error saving https://github.com/spf13/viper to cache - Error: open /root/.glide/cache/info/https-github.com-spf13-viper.json: no such file or directory
```

vs

```
[DEBUG] Attempting to find current branch for https://gopkg.in/logfmt.v0
[DEBUG] Saving default branch for https://gopkg.in/logfmt.v0
[DEBUG] Unable to cache default branch because caching is disabled
[DEBUG] Attempting to find current branch for https://gopkg.in/stack.v1
[DEBUG] Saving default branch for https://gopkg.in/stack.v1
[DEBUG] Unable to cache default branch because caching is disabled
[DEBUG] Attempting to find current branch for https://github.com/hashicorp/hcl
[DEBUG] Saving default branch for https://github.com/hashicorp/hcl
[DEBUG] Unable to cache default branch because caching is disabled
```

[...]

PR #265, opened 2016-02-17
Do not try to cache branch if caching is disabled

mhoglan:
Part of the change for #264 
Addresses the skipping branch caching step if caching is disabled.  

Still needs fix for global cache disabled flag in `repo/cache.go`

### DIFF
--- a/repo/vcs.go
+++ b/repo/vcs.go
@@ -421,24 +421,25 @@ func VcsGet(dep *cfg.Dependency, dest, home string, cache, cacheGopath, useGopat
 	gerr := repo.Get()
 
 	// Attempt to cache the default branch
-	branch := findCurrentBranch(repo)
-	if branch != "" {
-		// we know the default branch so we can store it in the cache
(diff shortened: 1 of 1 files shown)


mattfarina:
I think there's an issue ensuring the cache directory is properly in place, too. I'll look at this and the whole issue in a day or two. Would like to get this whole thing fixed for 0.9.1 release.

mhoglan:
I did look at that some too, the `EnsureCacheDir` was commented out and there is no where else that tries to create the cache directory.  The reason the cache directory ends up being created when you turn it on is the creation of the `json` to put in the cache in the `VcsGet` function causes the whole path to get created.  

from `cache.go#L90`
```
	p := filepath.Join(location, "cache", "info", key+".json")
	f, err := os.Create(p)
```

mattfarina:
This looks good. I'll fold it into 0.9.1 and, if someone else doesn't merge first, I'll merge it in on Tuesday.

PR #266, opened 2016-02-17
Add `brew` command to aid creating homebrew formulas

heewa:
Manually creating Homebrew "resource" definitions for a formula is pretty tedious and error prone, especially keeping them correctly updated over time. This change adds `glide brew` to convert the lockfile to homebrew resource definitions.

I also factored out some of the logic I needed into the `cfg` package, but I kept those changes in two separate commits, for clarity.

### DIFF
--- a/action/brew.go
+++ b/action/brew.go
@@ -1,0 +1,88 @@
+package action
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/Masterminds/glide/cfg"
+	"github.com/Masterminds/glide/msg"
+	gpath "github.com/Masterminds/glide/path"
+)
+
+// Brew converts dependencies to Homebrew resources and prints them, using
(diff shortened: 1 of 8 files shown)


mattfarina:
First, I absolutely love that you created a brew command. That's the kind of command I would personally use as I put stuff in Homebrew.

But, this is the kind of thing that belongs in a plugin rather than a core part of Glide. A lot of people won't use it and it's not required for package management. Is that something you'd consider doing?

If so we'd like to link to it as an example of a plugin. Would love to tell people about it.

heewa:
Oh, I didn't know about plugins. I'll do that instead. Thanks!

heewa:
Pending changes from PR #489, here's the plugin version of this change: [glide-brew](https://github.com/heewa/glide-brew)

PR #269, opened 2016-02-18
make glide work without use go vendor

hanyong:
I used glide and think it's great!
but it REQUIRES the use of GO VENDOR feature, 
and some of my tool chain,  such as gocode, goclipse,  doesn't support go vendor well.

So I think it well be greate if glide can work without use the go vendor feature (just similar the lower version of it does).
for the minimalist modification, I just do

1. Check if go vendor feature is enabled, if it's not, do my tricks as fellow. (I just Comment out the `os.Exit(1)` statement from `EnsureGoVendor`.)
2. make "vendor" as a symlink to the "src" directory of some GOPATH, in my case, it's the first GOPATH which ends with "/_vendor". I use gopath because I want the "_vendor" directory be out of the my source tree.
3. delete the "vendor" symlink after "install/update/get" command, so I will not confused for it.

I disabled "GO15VENDOREXPERIMENT" feature, and tested the code from my ubuntu PC,  it looks works well.
if the "GO15VENDOREXPERIMENT" feature is enabled, every thing works as the original logic,

I think the change not only DOSE NOT hurt people who use the go vendor featured,
but also helps people, like me, who do not use the go vendor feature very much.
so I created this pull request, and loyal hope glide could work without go vendor in the future,
best in a more elegant way.

### DIFF
--- a/action/ensure.go
+++ b/action/ensure.go
@@ -42,22 +42,25 @@ func EnsureCacheDir() {
 }
 
 // EnsureGoVendor ensures that the Go version is correct.
-func EnsureGoVendor() {
+func EnsureGoVendor() bool {
 	// 6l was removed in 1.5, when vendoring was introduced.
 	cmd := exec.Command("go", "tool", "6l")
 	if _, err := cmd.CombinedOutput(); err == nil {
 		msg.Warn("You must install the Go 1.5 or greater toolchain to work with Glide.\n")
-		os.Exit(1)
+//		os.Exit(1)
+		return false
 	}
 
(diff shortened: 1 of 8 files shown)


xentek:
For what it's worth, go1.7 will remove the environment variable and the `vendor` feature may not have a way to be disabled... meaning that this change won't last for much longer than 6 months when 1.7 comes around...

hanyong:
Seems new guru tool support go vendor good, I agree that work without go vendor feature is useless.

PR #278, opened 2016-02-23
Fixed #274: Was fetching dependencies of top level repo even when not imported

### LINKED ISSUE #274: only get dependencies of given subpackage not working
According to the release notes, since 0.9 only dependencies of specified subpackages should be retrieved recursively, to reproduce:

```
cd /tmp
mkdir glide-test
glide get github.com/hashicorp/terraform/dag
```
instead of  downloading the dependencies of the dag package it will download all dependencies of terraform. Is the because of Godeps being used? Any way to overwrite the behaviour?

mattfarina:


dmitris:
that seems to work - tested on https://github.com/dmitris/deptest4glide as well as our internal projects that don't build with the current master version because of dependencies' handling.

Would you consider cutting a 0.9.1 release to address `Glide now looks, by default, at the import tree in use rather than the complete possible tree of imports`, please?  Thanks.

mattfarina:
@dmitris A 0.9.1 will be coming within the next day barring something unforeseen happening.

PR #305, opened 2016-03-05
Fixes 304: glide tries to update ignored packages

fschl:
`resolveImports` did not check for `.HasIgnore` when building the final package list to update/install.


technosophos:
Good catch @fschl! LGTM

@mattfarina Can we merge this big fix for 0.9.2?

PR #306, opened 2016-03-05
Don't call SetVersion() twice

thockin:
While understanding the output of the tool, I noticed this.


technosophos:
LGTM. @mattfarina can we merge this for 0.9.2? It's a minor bug fix.

PR #307, opened 2016-03-05
Make 'already set' be a debug message

thockin:
This is a big reduction in non-useful noise when running.


technosophos:
LGTM.  That was my fault.

PR #309, opened 2016-03-06
Track updated packages to avoid dup work

### LINKED ISSUE #308: fetches deps more than once
We have a dep ox X.  Something else we depend on has a dep on X, too.  When glide runs I get duplicate "Fetching updates for X".  One fix is to keep a map[string]bool that tracks updated packages.  To pass that through into both of the places that call VcsUpdate() is a little messy (ConcurrentUpdate() and InVendor() both call it).  As a hack, I added it to cfg.Config and plumbed it around.  It does seem to work but it's not the most clean.

Thoughts?

thockin:
Fixes #308 

I am not really happy with this, but I thought it was a better start to avoid passing around a whole additional argument to all the actions.  Better ideas welcome.

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -48,6 +48,17 @@ type Config struct {
 	// DevImports contains the test or other development imports for a project.
 	// See the Dependency type for more details on how this is recorded.
 	DevImports Dependencies `yaml:"devimport,omitempty"`
+
+	// Updated tracks which packages have already been updated, so we don't
+	// duplicate work.
+	Updated map[string]bool `yaml:"-"`
+}
+
+// NewConfig allocates a Config structure with all fields properly initialized.
+func NewConfig() *Config {
+	cfg := &Config{}
+	cfg.Updated = map[string]bool{}
+	return cfg
 }
 
 // A transitive representation of a dependency for importing and exporting to yaml.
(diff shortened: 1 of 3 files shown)


technosophos:
Ah! Now I see what you were talking about in the issue. I had mentally confused VCS gets and dependency resolution. Really, there shouldn't be a case where a VcsGet operation is even requested twice, so there is some case in dependency resolution where we are not catching this case and it is trickling to the VcsGet layer. I'll be looking for that.

Thanks again!

@mattfarina I'd vote for putting this into 0.9.2 as well. It does clearly fix a bug that is clearly a performance issue.

mattfarina:
I like where this is going but I don't like that this information was added to the config object. I'd like to avoid those becoming a junk drawer. I have a couple ideas to clean this up.

thockin:
I had the same feeling, but it was that or literally pass a 2nd arg
everywhere conf was passed. Open to ideas. I am not attached to this, so
feel free to close and replace it if you get there before me.

mattfarina:
@thockin I like what you started. I'm in the process of building on in.

thockin:
Another alternative would be just to sleep a private may in vcs-management
code...

mattfarina:
I'm going to close this because of #315 (which builds on this code). Details in that pull request.

PR #311, opened 2016-03-06
Don't process deps if 'get' was a no-op

thockin:
This short-circuits a 'get' if there was nothing to actually get.


technosophos:
LGTM. @mattfarina if you're good with this, can we merge in time for 0.9.2?

PR #315, opened 2016-03-07
Track updated packages to avoid dup work

mattfarina:
This does a few things:

* Builds on #309 (which can now be closed).
* Handles map writing in a concurrency safe manner. As of Go 1.6 the runtime will panic if maps are concurrently written/read from.
* Moves the updated tracking from the config instance to the installer.

### DIFF
--- a/repo/installer.go
+++ b/repo/installer.go
@@ -109,8 +109,8 @@ func (i *Installer) Install(lock *cfg.Lockfile, conf *cfg.Config) (*cfg.Config, 
 		return newConf, nil
 	}
 
-	ConcurrentUpdate(newConf.Imports, cwd, i)
-	ConcurrentUpdate(newConf.DevImports, cwd, i)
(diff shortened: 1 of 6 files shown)


repo/installer.go
```diff
@@ -244,7 +254,7 @@ func ConcurrentUpdate(deps []*cfg.Dependency, cwd string, i *Installer) error {
 				select {
 				case dep := <-ch:
 					dest := filepath.Join(i.VendorPath(), dep.Name)
-					if err := VcsUpdate(dep, dest, i.Home, i.UseCache, i.UseCacheGopath, i.UseGopath, i.Force, i.UpdateVendored); err != nil {
+					if err := VcsUpdate(dep, dest, i.Home, i.UseCache, i.UseCacheGopath, i.UseGopath, i.Force, i.UpdateVendored, i.Updated); err != nil {
```
technosophos: I guess at some point we should investigate having VcsUpdate take an Installer as an arg.

technosophos:
Looks good to me.

Issue #318, opened 2016-03-08
Strip vcs data


mattfarina:
Add an option `--strip` flag to remove VCS directories from the `vendor/` folder.

There are two options for this.

1. Remove all VCS directories found in the `vendor/` folder.
2. Limit it to just those in projects managed by Glide.

In the #2 case if there is another dependency in the `vendor/` folder that Glide doesn't know about the VCS directory would remain.

Thoughts?

technosophos:
I'm not sure there is a clear distinction between "managed" and "unmanaged" repositories.

Say I manually vendor "foo", and then use `glide get` to vendor "bar", but "bar" depends on "foo". For all intents and purposes, Glide now treats "foo" as a managed dependency simply because of "bars" transitive dependency. It will now automatically do all the other Glide-y operations on "foo". In fact, the only dependencies that Glide won't de facto manage are the ones that are simply not imported at all.

So effectively, either route will get you the same outcome. You might as well start with the easy case and just do 1.

LK4D4:
I agree that 1 is simpler. And I think it's broken use case to use different vendor strategies within single project.

mattfarina:
Folks, can you take a look at #325. It attempts to implement this feature.

kron4eg:
`glide list` appears to be broken now.

here is my yaml and lock files

```
package: .
import:
- package: github.com/boltdb/bolt
- package: github.com/rogpeppe/godef
  repo: https://github.com/abourget/godef
  vcs: git
  version: feature/go15-vendor-experiment
```

```
hash: 5ffeacb33c685c181509d3e0058a7f7a8e0fa5a7be85309ab2f3e6a25bcc2787
updated: 2016-03-12T14:00:50.961900683+02:00
imports:
- name: github.com/boltdb/bolt
  version: 0fd4c0547d204c7b1cad6db6f3adad5f2cf453e5
- name: github.com/rogpeppe/godef
  version: 60e71fcb82867b2b0ea18d4f5625a3ec9720b5e2
  repo: https://github.com/abourget/godef
  vcs: git
- name: golang.org/x/sys
  version: 7a56174f0086b32866ebd746a794417edbc678a1
  subpackages:
  - unix
devImports: []
```

and here is outputs of different commands:

```
$ glide list
[INFO] Package golang.org/x/sys/unix found in vendor/ folder
[WARN] Version not set for package golang.org/x/sys/unix
[INFO] Sorting...
INSTALLED packages:
[...]

PR #325, opened 2016-03-09
Implements #318: Opt-In to strip VCS metadata from vendor directory

mattfarina:


technosophos:
Man... this makes me nervous... but I think we should do it. I'd like to get it merged in ASAP so we can get lots of tire kicking before the release.

Code LGTM.

glide.go
```diff
@@ -404,6 +404,10 @@ Example:
 					Name:  "use-gopath",
 					Usage: "Copy dependencies from the GOPATH if they exist there.",
 				},
+				cli.BoolFlag{
+					Name:  "strip",
```
thockin: my only nitpick would be that the name could be more obvious.  --rmvcs or --strip-vcs or something would be clearer, but I'm not looking a gift horse in the mouth.

mattfarina: I like the name `strip-vcs`.

Issue #320, opened 2016-03-08
Resolve devImports


mattfarina:
In addition to imports, Glide should be able to resolve imports in test files.

IvanMalison:
Status?

mattfarina:
@IvanMalison A PR was started in #331. It still needs to be finished.

mattfarina:
I'm moving this to the 0.11 release since it's not complete. The work has already been started.

mattfarina:
Implemented on the tip of master.

Issue #168, opened 2015-12-15
Allow glide.yaml to declare dependencies as "development-time only"


xeger:
It's common for a package to depend on third-party code in order to run tests or perform other development work on the package itself, but not to depend on that code at runtime. The canonical example would be a library like ginkgo or omega: I may choose to write my tests using ginkgo, but if someone imports my package to use my exports in their own executable, they do _not_ need ginkgo.

Because glide honors any `glide.yaml` it finds in transitive dependencies, there is no way for me to use glide both to manage my project's development-time dependencies _and_ provide a dependency specification to others who are importing my package. Since I want others to be able to use my code, I am forced to omit development dependencies from `glide.yaml` and manage them by hand; ouch!

I'd like to add the capability to glide to mark certain dependencies as "local," i.e. not to be considered when one's package is imported into another person's project. There are many ways to achieve this, so I thought I would gather some feedback before coding anything.

Some sample ideas for supporting conditional dependencies:

- Introduce a flag such as `development: true|false` to each dependency listed in glide.yaml
- Introduce a more flexible concept such as "tagging" of dependencies, and allow callers of the`glide` CLI to provide a whitelist or blacklist of dependencies that should be considered in building the graph
- Introduce a separate file i.e. `glide-local.yaml` that glide ignores unless it lives in initial working dir.

mattfarina:
There is a `devImports` option in the `glide.yaml` for this that doesn't actually work at the moment. It's not wired up. It's currently the next most highly requested feature so we will be implementing it in the near future.

technosophos:
I am thinking that probably what we will do is add a `--dev` flag to the relevant commands, and that will read the `devImports` section of a Glide file as well as the regular `imports` section.

I'm wondering if we should have a separate lock file for dev dependencies.

xeger:
Thank you; a `devImports` plus an opt-in `--dev` flag would fit the bill quite nicely. I stand ready to test -- I'd offer to help implement but it sounds like you're moving quickly and I'm not up to speed on your project yet, so I'll be a guinea pig instead.

pspeter3:
Is there an open issue that is tracking this development?

mattfarina:
@pspeter3 yes. See #152 and #103. We're currently cleaning up the internals of Glide which will remove some old code and make it easier for people to contribute. Once that is done this is one of the higher priority items. It will be coming soon since much of that cleanup work is done.

Issue #103, opened 2015-10-12
glide ignore packages in test files


hvnsweeting:
when I run glide guess/glide pin, they showed list of packages but missing those that I use for *_test.go.
Is that a feature or should it be fixed?

Thanks

mattfarina:
I had not thought about this before. What's happening here is the `go/build` package has two separate things for imports. `Imports` is for your code and `TestImports` is for any imports in test code. We are just using the former and end up missing test imports.

Should test imports be on the `Imports` list in the yaml file or should they be on `devImports`?

@technosophos any thoughts on this?

technosophos:
My initial intent when I added `devImports` was for test and build dependencies.

davecgh:
I see the opposite with https://github.com/syndtr/goleveldb.  There are several dependencies that are only needed to run the tests which are included in the `imports` section of the lock file.

Glide should gain a `-t` flag (or an aptly named option in the yaml) just like `go get` has in my opinion.  Specifically:

```
The -t flag instructs get to also download the packages required to build
the tests for the specified packages.
```

One of the primary purposes of glide is, of course, to provide reproducible builds and installs to end user and developers alike.  To that end, testing of dependencies is squarely aimed more at developers than end users and thus should really be opt-in with a flag/optional keyword, just like `go get` does it.  A normal end user will typically want to the produce the final stable binary without needing or wanting to download the additional packages only needed for testing, particularly of dependencies.

It is slightly debatable whether the test-only dependencies of the project itself are included by default, but packages only required by the tests in dependencies should most definitely not be.

krancour:
I am running into this as well.  I have a project of _just_ integration tests written in go.

My workaround has been to include a dummy non-test file that looks like this:

```
package tests

import (
	// This is a workaround for the fact that glide seems not to recursively resolve packages needed
	// only for testing.
	_ "k8s.io/kubernetes/pkg/client/unversioned"
)

```

But I'd rather not have to do that, obviously.

mattfarina:
I've added #320 to track resolving test imports.

mattfarina:
Please see #461 which is trying to fix the management of test imports. Reviews appreciated.

mattfarina:
Fixed. Tip of master now handles test imports.

PR #331, opened 2016-03-11
Resolve devImports

alde:
Can be omitted by using the new flag --without-dev

glide update [--without-dev]
glide install [--without-dev]

### DIFF
--- a/action/get.go
+++ b/action/get.go
@@ -58,7 +58,7 @@ func Get(names []string, installer *repo.Installer, insecure, skipRecursive bool
 	}
 
 	// Set Reference
-	if err := repo.SetReference(confcopy); err != nil {
+	if err := repo.SetReference(confcopy, installer); err != nil {
 		msg.Err("Failed to set references: %s", err)
 	}
 
@@ -84,7 +84,7 @@ func writeLock(conf, confcopy *cfg.Config, base string) {
 	if err != nil {
 		msg.Die("Failed to generate config hash. Unable to generate lock file.")
 	}
-	lock := cfg.NewLockfile(confcopy.Imports, hash)
+	lock := cfg.NewLockfile(confcopy.Imports, confcopy.DevImports, hash)
 	if err := lock.WriteFile(filepath.Join(base, gpath.LockFile)); err != nil {
 		msg.Die("Failed to write glide lock file: %s", err)
 	}
--- a/action/install.go
+++ b/action/install.go
@@ -48,7 +48,7 @@ func Install(installer *repo.Installer) {
 	msg.Info("Setting references.")
 
 	// Set reference
-	if err := repo.SetReference(newConf); err != nil {
+	if err := repo.SetReference(newConf, installer); err != nil {
 		msg.Err("Failed to set references: %s (Skip to cleanup)", err)
 	}
 
--- a/action/remove.go
+++ b/action/remove.go
@@ -25,9 +25,9 @@ func Remove(packages []string, inst *repo.Installer) {
 	// Copy used to generate locks.
 	confcopy := conf.Clone()
 
-	confcopy.Imports = inst.List(confcopy)
+	confcopy.Imports, confcopy.DevImports = inst.List(confcopy)
 
-	if err := repo.SetReference(confcopy); err != nil {
+	if err := repo.SetReference(confcopy, inst); err != nil {
 		msg.Err("Failed to set references: %s", err)
 	}
 
--- a/action/update.go
+++ b/action/update.go
@@ -29,7 +29,7 @@ func Update(installer *repo.Installer, skipRecursive bool) {
 
 	// Set the versions for the initial dependencies so that resolved dependencies
 	// are rooted in the correct version of the base.
-	if err := repo.SetReference(conf); err != nil {
+	if err := repo.SetReference(conf, installer); err != nil {
 		msg.Die("Failed to set initial config references: %s", err)
 	}
 
@@ -52,7 +52,7 @@ func Update(installer *repo.Installer, skipRecursive bool) {
 		// installer set them as it went to make sure it parsed the right imports
 		// from the right version of the package.
 		msg.Info("Setting references for remaining imports")
-		if err := repo.SetReference(confcopy); err != nil {
+		if err := repo.SetReference(confcopy, installer); err != nil {
 			msg.Err("Failed to set references: %s (Skip to cleanup)", err)
 		}
 	}
@@ -77,7 +77,7 @@ func Update(installer *repo.Installer, skipRecursive bool) {
 		if err != nil {
 			msg.Die("Failed to generate config hash. Unable to generate lock file.")
 		}
-		lock := cfg.NewLockfile(confcopy.Imports, hash)
+		lock := cfg.NewLockfile(confcopy.Imports, confcopy.DevImports, hash)
 		if err := lock.WriteFile(filepath.Join(base, gpath.LockFile)); err != nil {
 			msg.Err("Could not write lock file to %s: %s", base, err)
 			return
--- a/cfg/lock.go
+++ b/cfg/lock.go
@@ -82,11 +82,12 @@ type Lock struct {
 }
 
 // NewLockfile is used to create an instance of Lockfile.
-func NewLockfile(ds Dependencies, hash string) *Lockfile {
+func NewLockfile(ds Dependencies, devs Dependencies, hash string) *Lockfile {
 	lf := &Lockfile{
-		Hash:    hash,
-		Updated: time.Now(),
-		Imports: make([]*Lock, len(ds)),
+		Hash:       hash,
+		Updated:    time.Now(),
+		Imports:    make([]*Lock, len(ds)),
+		DevImports: make([]*Lock, len(devs)),
 	}
 
 	for i := 0; i < len(ds); i++ {
@@ -101,7 +102,20 @@ func NewLockfile(ds Dependencies, hash string) *Lockfile {
 		}
 	}
 
+	for i := 0; i < len(devs); i++ {
+		lf.DevImports[i] = &Lock{
+			Name:        devs[i].Name,
+			Version:     devs[i].Pin,
+			Repository:  devs[i].Repository,
+			VcsType:     devs[i].VcsType,
+			Subpackages: devs[i].Subpackages,
+			Arch:        devs[i].Arch,
+			Os:          devs[i].Os,
+		}
+	}
+
 	sort.Sort(lf.Imports)
+	sort.Sort(lf.DevImports)
 
 	return lf
 }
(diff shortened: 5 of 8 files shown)


mattfarina:
Note, this implements #320.

mattfarina:
@alde This is a good start. Thanks. It still needs to have the resolver get the imports from the test files and put them into the devImports.

alde:
Cheers, I'll try to fix that this weekend, or on monday.

technosophos:
I was originally going to suggest that we do not include dev dependencies by default (e.g. that we have `--include-dev` instead). But as I thought about the use case, it feels to me like the cases where we'd not want dev imports are for CI/CD and stuff like that. So I agree with this choice.

When we encounter Glide.yaml files in dependencies, do we obey this flag in regards to those? I'm kinda thinking no. But I could be swayed the other way.

alde:
My idea with --without-dev was that if you know that you won't be running the tests for example you'd know that you don't want/need them installed

I am however having some issues with the resolver, as in I'm not sure I fully understand what it's meant to do. I assumed at first it went through the .go files and found imports, and as such it would need to go through the _test.go files for dev imports, but that doesn't seem to be the case

technosophos:
Maybe we should step back a moment and describe what `devImports` is for. I think @alde may be right: we may not need to go through the `_test.go` files to get what we want.

Here's my first pass:

- devImports is for packages that are required to develop _on the local project_ (not anything in `vendor/`)

Examples of devImports:

- An assert library that is used only in testing (for the local project), e.g. Ginkgo
- An external Go tool that is used for something like `go generate`

My assumption is that by default, when I import with `devImports`, I can do things like run `go test` on the project. But I am _not_ assuming that I'd be able to cd into vendor and run `go test ./...` and have that work.

Does that sound similar to what you were thinking @mattfarina and @alde ?

alde:
Yes, that was my thought, and the reason I started trying to patch it.

mattfarina:
@technosophos The external tools become a little more difficult to use for `devImports`. The generator is a great example. How do you detect it's a tool that should be installed and install it rather than just put it in `vendor`? I was considering scripts for the case. Then `devImports` can be things like imports only used in tests.

technosophos:
So based on the comments of @alde @mattfarina and me, it sounds like we're down to just this use case:

- devImports is for libraries necessary only for running tests on the present project (not on things in `vendor`)

Does that sound right?

lpetre:
I wanted to +1 the case for @technosophos's comment

> An external Go tool that is used for something like `go generate`

We do this currently for `fresh` and I would love for our dependency management solution to cover this use case. Otherwise, do I need to have a separate way (makefile, ant, etc) of telling my developers to `go get` a half-dozen packages at specific versions/urls/etc?

mattfarina:
@alde Is this something I should pickup and finish completing or is this something you wanted to finish rounding out?

alde:
Yeah, sorry, I've got it a bit hectic atm.

If you want to finish it, please feel free, I don't know when I'll have the time to focus on it again :(

ches:
> We do this currently for fresh and I would love for our dependency management solution to cover this use case. Otherwise, do I need to have a separate way (makefile, ant, etc) of telling my developers to go get a half-dozen packages at specific versions/urls/etc?

Ditto.

By comparison, I've been happy using `gb` on application-only projects, but am hoping Glide can take its place for projects with a library component so they can be `go get`-able. With `gb`, executables from vendored dependencies get built into `vendor/bin`, which is similar to how some analogs like Bundler in Ruby behave. My teams standardized on [direnv](http://direnv.net/) with a project `.envrc` like the following, so that the project's executables including vendored tools like `ginkgo` are readily available:

```sh
export GOPATH="$PWD/vendor:$PWD"
PATH_add "$PWD/vendor/bin"
PATH_add "$PWD/bin"
```

Of course `gb` eschews `GOPATH` entirely, so that allows for flexible answers to what to do about the executables (`GOPATH` in the `.envrc` is just to make tools like `gocode`, `goimports`, etc. play along with the project FWIW).

Would it be viable for Glide to install the `devImports` executables in `vendor/bin` (and possible without gross hacks around what `go/build` wants to do naturally—I'm not familiar with the internals)? It seems reasonably convenient for users to invoke project-local tools from there without needing anything extra like direnv, and there's some similar precedent, but I'm not sure if it would run afoul of any current or foreseeable assumptions/conventions of Glide, `go` tooling, or others…

I start to wish that Go vendor experiment had followed `vendor/{bin,pkg,src}` like `gb` does…

ches:
> I was originally going to suggest that we do not include dev dependencies by default (e.g. that we have --include-dev instead). But as I thought about the use case, it feels to me like the cases where we'd not want dev imports are for CI/CD and stuff like that.
>
> ...
> - devImports is for libraries necessary only for running tests on the present project (not on things in `vendor`)
>
> Does that sound right?

I think that a critical conflation of dev versus test has pervaded the thread—I was first confused by @technosophos's quoted mention of CI. I thought, "I certainly do *not* want to use `--without-dev` on CI if Ginkgo is in `devImport`", but @lpetre's example of `fresh` is exactly a case where you would. (Though, it's not really an "import", you don't import `fresh` in project code—maybe `devPackages`?).

Ruby's Bundler supports arbitrary groups with dev and test being common convention; all are installed by default, and typically `--without=dev` is used on CI and `--without=dev,test` in production. This is akin to possibilities @xeger proposed in #168.

But presumably Go build tooling does a fine job already of keeping test libraries from bloating production artifacts when they're only imported in `_test.go` files. So it seems to me like `devImports` is all that is really needed (arbitrary groups/tags seem overkill for now), but *test* dependencies do *not* go there but in `import` instead (and something must be worked out about executables like the `ginkgo` runner, as well as dev executables like `fresh`, as discussed earlier).

In that case I think it becomes the clear choice that `devImports` is only considered for the local project and not recursively within `vendor`, as @technosophos's first version above suggested. Thus making the option an opt-out `--without-dev` seems more convenient for the common use case to me.

The downside is that, if Glide looks at `glide.yaml`/`lock` files recursively in dependencies, it would be ideal if it didn't just install everything in `import`, it should probably filter out deps that it discovers are only imported in `_test.go` files. @davecgh made the fair point on #103 that `go get` doesn't waste time and space fetching test dependencies by default, unless you use `-t`. Maybe it would be easier to additionally have `testImport`, then? Doesn't help avoid waste for deps that don't use Glide, though, if test imports aren't already being omitted there (I'm not sure). Personally I don't care if `go test ./...` errors in the default case, I'm practically never going to use that without `glide novendor` so I support the opt-in `-t` equivalent.

mattfarina:
Here's a thought, should it change from `devImports` to `testImports` to clear up some confusion?

sdboyer:
@mattfarina I think that renaming would probably be helpful. I may rejigger the corresponding naming for `vsolver`, too.

@ches:
> The downside is that, if Glide looks at `glide.yaml/lock` files recursively in dependencies, it would be ideal if it didn't just install everything in `import`, it should probably filter out deps that it discovers are only imported in `_test.go` files.

Yeah, this is part of a more general problem - what @technosophos and I have been calling the "bi-modal analysis" problem: do you operate at the project or package level in building the depgraph? When you do the former, you may get excess packages; when you do the latter, you have no version information. The former is wasteful, but ultimately correct; the latter is just wrong (which is why we're all here).

It's not intractable, I don't think; it just involves making some tradeoffs. However, I've deferred work on it in favor of trying to get `vsolver` integration into glide going ASAP in #384. Once that's in, though, tackling that general problem - which would hopefully take care of this along with - is basically next on my TODO list.

ches:
> Here's a thought, should it change from `devImports` to `testImports` to clear up some confusion?

@mattfarina If we've resolved that the semantic is meant to be analogous to `go get -t`, then yes I think that makes sense and focuses the discussion (and it sounds like it's pretty much what @alde's PR was conceived as, modulo naming).

Consideration of a feature for dev tooling like the `go generate` support and `fresh` examples could be cleanly moved to a subsequent ticket, it seems.

arvenil:
Hi guys,

What's the status of that?

The lack of vendoring imports from tests is pretty big issue for me, as basically you can't run tests that relay on some packages that aren't required in the code itself. To bypass the issue right now I do `cp somepackage_test.go somepackage_x.go; glide up; rm somepackage_x.go`.

Btw this was working couple months ago, as then glide was fetching ALL dependencies... I love that it doesn't do that anymore but yeah since then it doesn't parse dependencies from test files:/

arvenil:
Something like this (without this patch) fixes the issue for me
```go
index dc88f62..62b4731 100644
--- a/dependency/resolver.go
+++ b/dependency/resolver.go
@@ -266,7 +266,7 @@ func (r *Resolver) ResolveLocal(deep bool) ([]string, error) {
 				return err
 			}
 		} else {
-			imps = p.Imports
+			imps = append(p.Imports, p.TestImports...)
 		}
 
 		// We are only looking for dependencies in vendor. No root, cgo, etc.
```

I've started to think that maybe a good idea would be to merge just that one line to fix the issue with `_test.go` deps not being vendored... and then, in next versions, you could guys improve that with that e.g. by separating app and test deps. I don't see a drawback of fixing the issue first, and then working on improvements. Fixing this like that right now doesn't collide with moving later deps from `imports` to `devImports` in yaml

mmindenhall:
Please note that it's also possible for test files in package `foo` to declare their package as `foo_test` (from `go help test`):

> Test files that declare a package with the suffix "_test" will be compiled as a
> separate package, and then linked and run with the main test binary.

If the list of test imports is being collected from the `Package` struct, the entire set of imports would include:

```go
	imps := p.Imports                      // imports from GoFiles, CgoFiles
	imps = append(imps, p.TestImports...)  // imports from TestGoFiles
	imps = append(imps, p.XTestImports...) // imports from XTestGoFiles
```

TommyM:
Is there any news on this PR or similar functionality (vendoring test dependencies) going into glide anytime soon?

sdboyer:
@TommyM yes, this problem is already generally solved in the new engine (unconditionally, though - allowing a flag that would skip installing non-test imports is another can of worms).  

so, just gotta get that in :) #384

kevin-cantwell:
I know this is closed, but perhaps this will be helpful for those trying to vendor their test deps with glide. It's a crazy stupid long one liner in bash that passes all recursive test deps into `glide get`:

```
( go list -f '{{join .TestImports "\n"}}' $(glide novendor) | xargs go list -f '{{join .Deps "\n"}}'  | xargs go list -f '{{if not .Standard}}{{.ImportPath}}{{end}}' ; go list -f '{{join .XTestImports "\n"}}' $(glide novendor) | xargs go list -f '{{join .Deps "\n"}}'  | xargs go list -f '{{if not .Standard}}{{.ImportPath}}{{end}}' ) | xargs glide get
```

Unfortunately, this only works if your vendor directory is empty. Otherwise you'll end up with a bunch of subpackages in your `glide.yaml` with the "vendor/" prefix.

mattfarina:
Early next week I'll have a PR to more formalize this. The hard part is already done.

PR #332, opened 2016-03-11
Fixed update: remove ignored dependencies

timstudd:
Without this change, ignored dependencies are written to glide.lock.  This is problematic for local sub-package imports, for which a VCS cannot be determined.  When explicitly ignoring local sub-packages, glide.lock should not include them.

### DIFF
--- a/action/update.go
+++ b/action/update.go
@@ -104,5 +104,4 @@ func removeIgnoredImports(conf *cfg.Config) {
 			}
 		}
 	}
-	return
 }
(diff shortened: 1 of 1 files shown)


mattfarina:
@timstudd Thanks for bringing this up. It's definitely a bug.

Two things:

1. Did you know there is a [`conf.HasIgnore()`](https://godoc.org/github.com/Masterminds/glide/cfg#Config.HasIgnore) method?
2. Ignored packages should not be on `confcopy`. That earlier bug should be found and squashed.

Do you want to take a shot at finding why ignored packages are ending up on `confcopy` or shall I look into it?

timstudd:
@mattfarina I found that the issue was due to my local environment.  I'm using OS X and had symlinks with both GOROOT and GOPATH to a outside directory.  This apparently caused the issue as removing the symlink from GOROOT resulted in glide.yaml no longer including the local sub-packages.  Thanks for your consideration, but I'll close this PR.

Issue #297, opened 2016-03-01
Glide Update is Unusable on Windows


kumarharsh:
I have a project with this structure:

```
| my-go-project
---| frontend
    ---| node_modules
---| vendor (glide packages go here)
---| glide.yaml
---| glide.lock
```

Whenever I do `glide update`, glide tries, for some baffling reason, to go into my node_modules folder right down to the absolute bottom of each node_module, and then dies because of running into the `MAX_PATH` problem in windows, and throws this error:

> [ERROR] Failed to build an initial list of packages to scan: GetFileAttributesEx D:\Workspace\<my-project-directory>\<frontend>\node_modules\babel\node_modules\chokidar\node_modules\anymatch\node_modules\micromatch\node_modules\braces\node_modules\expand-range\node_modules\fill-range\node_modules\is-number\LICENSE: The system cannot find the path specified.

And hence, it never installs any new package - basically it's pointless having it.

mattfarina:
I can explain what's happening.

Glide, like the `go` tools, scans the file system to figure out the dependencies. This was a feature request some time ago and it's an expectation because tools like `go get` do it as well.

This issue may be due to the Windows path length issues that come up in node.js with the deep nesting. Is that a problem? What version of npm are you using?

mattfarina:
Glide walks the complete tree to be able to fetch all the dependencies. Glide creates reproducible packages and versions. Go projects can have go files in nested subdirectories to it needs to scan for them.

Glide can ignore some packages now. We likely need to add the ability to ignore some directories. That would solve your issue.

kumarharsh:
Thanks. I understand that that's how go works, but if we are "absolutely" sure that no go file can be in some directories, it might be useful to save that trouble (and time) on every `glide update`

kyteague:
Why does glide update look at all of the go files again when update is ran?

This is really non-standard behavior for package mangers.  I understand looking through the files to generate the initial glide.yaml, but after that it creates a lot of headaches.  In fact, I believe this behavior is the cause of most issues on GitHub.

I really want to use glide, but it just behaves in unpredictable ways and I have spent way too much time getting it to work (and still doesn't).

kumarharsh:
@mattfarina any updates on this?

mattfarina:
@kumarharsh I've personally got this as a use case so I'm working on it this week. Right now I'm attempting to figure out the new key for the `glide.yaml` file.

mattfarina:
[...]

PR #334, opened 2016-03-15
Issue #297: Adds exclude property for directories in local codebase to exclude from scanning.

mattfarina:


technosophos:
Yeah, this looks good to me. I'm slightly worried about the UX, since it's a little ambiguous what `exclude` means in comparison to `ignore`. I guess we _could_ call it `excludeDirs`, but I honestly prefer simplicity in naming.

kumarharsh:
yeah, maybe `excludeDirs` would be a better name to avoid confusion. But it's great nonetheless, I can actually use glide now! :confetti_ball:

mattfarina:
I'm not really happy with the name `exclude` but `ignore` was more useful before there was import tree walking. `ignore` is far less useful now.

The exclude name is used other places such as:
* [In Composer](https://getcomposer.org/doc/04-schema.md#exclude-files-from-classmaps) where it's `exclude-from-classmap`.
* Cargo (Rust) uses [`exclude`](http://doc.crates.io/manifest.html#the-exclude-and-include-fields-optional) though the context is a little different. It's for package files to send to the central repo.
* NPM uses and ignore file and is similar to Cargo in intent.

I'm going to update to `excludeDirs` to avoid confusion. It will happen and I'd prefer to answer fewer questions later.

PR #337, opened 2016-03-16
Print the path a recursive deps file was found in

thockin:
This is imperfect - it says:

```
Found Godeps.json file in /home/thockin/tmp/glide/src/k8s.io/kubernetes/vendor/github.com/opencontainers/runc
```

when it would be better to say:

```
Found Godeps.json file in vendor/github.com/opencontainers/runc
```

or:

```
Found Godeps.json file in k8s.io/kubernetes/vendor/github.com/opencontainers/runc
```

or:

```
Found Godeps.json file in vendored github.com/opencontainers/runc
```

Better fixes would be welcome instead of this, but this is better than nothing.

### DIFF
--- a/gb/gb.go
+++ b/gb/gb.go
@@ -24,7 +24,7 @@ func Parse(dir string) ([]*cfg.Dependency, error) {
 		return []*cfg.Dependency{}, nil
 	}
 
-	msg.Info("Found GB manifest file.\n")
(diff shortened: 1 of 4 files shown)


mattfarina:
Thanks. Merged and cleaned up the messages to be in the format:
```
Found Godeps.json file in vendor/github.com/opencontainers/runc
```

PR #351, opened 2016-03-24
Migrating chat to Gitter

mattfarina:
@technosophos shall we merge?


technosophos:
YAY! Yes, we should merge!

Issue #256, opened 2016-02-13
Add --porcelain to glide list


chancez:
It would be good to have something similar to `git status --porcelain` to `glide list` so the command can be used more programatically. My use case is to use the output of glide list to remove any packages not directly imported by my project from the vendor directory so when im vendoring I can have a more minimal vendor dir. 

The current issue is all the warnings and extra text that aren't go packages get printed and there's a lot of whitespace/indentation in the output, meaning I need to do some extra filtering and formatting before I can use it.

chancez:
This might deserve a new issue, but it would also be good if `glide list` had a flag so it could also inspect my project's test files imports, and include those in the list.

mattfarina:
@chancez I like the idea of making the output usable for other tools. I'm wondering if we should support a couple formats. Text by default but also a flag for json. Thoughts?

Note, we need to make sure the output goes to standard out, too.

chancez:
Yah having text and json would be awesome. Json could have lots of
metadata, one json object per line (event stream basically) and it could
easily be consumed by lots of tools.

PR #365, opened 2016-04-02
glide list: support json output format

chancez:
Sort of addresses #256. Might be worth having a format which spits out a newline delimited list, without the "INSTALLED/MISSING/GOPATH packages" headers.

### DIFF
--- a/action/list.go
+++ b/action/list.go
@@ -13,8 +13,7 @@ import (
 // Params:
 //  - dir (string): basedir
 //  - deep (bool): whether to do a deep scan or a shallow scan
-func List(basedir string, deep bool) {
-
+func List(basedir string, deep bool) PackageList {
 	basedir, err := filepath.Abs(basedir)
 	if err != nil {
 		msg.Die("Could not read directory: %s", err)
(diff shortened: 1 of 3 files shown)


technosophos:
I like this a lot.

mattfarina:
@chancez First, nice work on this. I like seeing this as json and json-pretty. I can use that myself.

I do have some feedback for changes (more on that in a moment). If you're not going to have time just let me know and I can make them via an additional commit.

glide.go
```diff
@@ -639,3 +649,42 @@ func glidefile(c *cli.Context) string {
...
 	return a
 }
+
+const (
+	textFormat       = "text"
+	jsonFormat       = "json"
+	jsonPrettyFormat = "json-pretty"
+)
+
+func outputList(l action.PackageList, format string) {
```
mattfarina: The `glide.go` file is already fairly long. Can we move `outputList` into `action/list.go`.  Maybe instead of `action.List` returning something it takes in the format switch and handles the display. That's how the other commands work.

glide.go
```diff
@@ -587,7 +588,16 @@ Example:
...
 			Action: func(c *cli.Context) {
-				action.List(".", true)
+				outputFormat := c.String("output")
+				pkgList := action.List(".", true)
+				outputList(pkgList, outputFormat)
+			},
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  "output, o",
+					Usage: "Output format. One of: json|json-pretty|text",
```
mattfarina: In the usage can you note the default format.

action/list.go
```diff
@@ -27,34 +26,29 @@ func List(basedir string, deep bool) {
...
-			msg.Puts("\t%s", pkg)
-		}
+	return PackageList{
+		Installed: installed,
+		Missing:   h.Missing,
+		Gopath:    h.Gopath,
 	}
 }
+
+type PackageList struct {
```
mattfarina: Can you add a comment describing what the type does. `golint` will bug us if you don't and then in the future we'll remember (if we forget).

mattfarina:
@chancez thanks for the contribution. I finished the tweaks off.

chancez:
@mattfarina Yeah, sorry, I was planning on finishing it off, but you were so quick, so I don't mind! Thanks.

PR #377, opened 2016-04-06
Fixes #371: Warn when name/location mismatch

### LINKED ISSUE #371: Why is glide vendoring the project that it's controlling
Right now I'm running `glide up` and glide is vendoring one of my subpackages? How and why is this happening?

mattfarina:
When the package in the glide.yaml does not match the current
location in the GOPATH warn the users.

### DIFF
--- a/action/ensure.go
+++ b/action/ensure.go
@@ -51,6 +51,8 @@ func EnsureConfig() *cfg.Config {
 			if name != conf.Name {
 				msg.Warn("The name listed in the config file (%s) does not match the current location (%s)", conf.Name, name)
 			}
+		} else {
+			msg.Warn("Problem finding the config file path (%s) relative to the current directory (%s): %s", b, cwd, err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


sdboyer:
:+1:

action/ensure.go
```diff
@@ -33,6 +35,25 @@ func EnsureConfig() *cfg.Config {
...
+	if err != nil {
+		msg.Err("Unable to get the current working directory")
+	} else {
+		// Determining a package name requires a relative path
+		b, err = filepath.Rel(b, cwd)
+		if err == nil {
+			name := buildContext.PackageName(b)
+			if name != conf.Name {
+				msg.Warn("The name listed in the config file (%s) does not match the current location (%s)", conf.Name, name)
+			}
```
technosophos: What if there is an error from Rel?

mattfarina: @technosophos An error should never happen. The reason I didn't do anything was I wanted to keep the user experience sane. I think I'll put a debugging message in.

Issue #372, opened 2016-04-04
Proposal: Ability to configure VCS rewrites


sectioneight:
So I realize this is a bit of a special case, but wanted to run by you to see if you'd be open to accepting a PR.

At my place of employment, we are switching to Glide (from Godeps). At the same time, we'd like to move away from vendoring. Glide is great for this, and we get reproducible builds **when GitHub is up**, but want to guard against repo renames/GH instability.

Internally, we have a read-through cache/mirror of GitHub repos, so that if we pull from a special git URI (e.g. `git clone git@mirror:github/Masterminds/glide`) it will read the latest state from GitHub, but still return you the code if GitHub goes down (via gitolite's mirroring options).

We'd like the ability to configure glide internally, such that if an engineer adds a github.com dependency to their glide.lock/yaml, it will get pulled through our mirror.

Alternatively, we are open to patching glide on the client side, such that any "glide get github.com/..." will result in the proper VCS settings in glide.yaml. This is slightly less preferable because then we have to distribute a patched glide.

I'm open to implementing this on an internal fork if you don't foresee others needing this functionality.

What I'm thinking is a `~/.glide.yaml` that has a block like:
```
vcs:
  rewrites:
    from: git@github.com
    to: git@mirror:github/
```

What do you think?

sdboyer:
I think you might be able to do this today, with the `repo` field that you can [declare on deps](https://github.com/sdboyer/vsolver/blob/master/glide.yaml). Ordinarily it's used to declare aliases/forks, but I think it could be bent to this purpose.

If `glide` was to support this properly - particularly for the use pattern you're describing - I'd argue it would be preferable to have it be something where the URIs are switched in transparently - not something declared in `glide.yaml`. Yes, that would probably entail compiling a custom binary...though glide does have a "plugin" system already - perhaps that could be leveraged.

Point is, having to repeatedly add that information in `glide.yaml` is not only likely to promote errors and inconsistencies because user input, but would also result in the default behaviors of commands like `glide get` becoming suboptimal for your workflow (users would need to go into the `glide.yaml` later to add that information). Best to avoid design choices that effectively cripple other parts of the tool, if possible.

PS - does your internal gitolite service *really* have more uptime 9's than github? :grin:

sectioneight:
It actually does, believe it or not :)

[...]

PR #380, opened 2016-04-12
Refactor installer / vcs creation & configuration

crsmithdev:
As discussed in #372 I am in the process of adding support for URL rewrites.  When looking at adding an additional configuration file (`.gliderc` or similar, TBD), I noted that it would be difficult to add in support for this under the current installer / vcs structure -- some vcs methods, for example, already have 9 parameters, and adding onto that is just going to pile up debt.

So, in preparation for doing that, I've refactored the installer / vcs code a bit to make creating and configuring them a bit more straightforward, and adjusted the method signatures accordingly.  This should make adding a new config file and passing config around much more straightforward.

I think there's certainly additional work to be done (further refactors, more testable abstractions, tests) but this seems like a good stopping point for review.


mattfarina:
@crsmithdev thanks for trying to clean this up. I had a chance to skim it (I'll give it a finer pass later).

From this one question came up. There are both `conf` and `confcopy`. The reason `confcopy` is passed into the installer is because transitive packages are added to this. We need both a pristine copy of conf and a version containing the transitive dependencies for doing things like writing the lock file.

Here is seems the installer skips confcopy (if not I might have missed it). That's going to create some issues.

Does that make sense?

crsmithdev:
Yes...that was a source of some confusion for me as well.  Would a workable approach be to replace the installer / vcs `Config` value, from `conf` to `confcopy`, after `confcopy` is created, and then swap it back @ the end of whatever function is creating the copy?  Or is there a better way, somehow?

technosophos:
I really like the config structs. Too much organic growth led to some icky function signatures. This is much cleaner.

crsmithdev:
I've updated this w/ more careful treatment of config copies.

repo/installer.go
```diff
@@ -49,15 +65,25 @@ type Installer struct {
...
+		Vcs: NewVcs(VcsOptions{
+			Home:           options.Home,
+			UseCache:       options.UseCache,
+			UseCacheGopath: options.UseCacheGopath,
+			UseGopath:      options.UseGopath,
+			UpdateVendored: options.UpdateVendored,
+			Force:          options.Force,
+			Vendor:         options.Vendor,
+			Config:         options.Config,
+		}),
```
mattfarina: Notice the `UpdateTracker` is attached to the `Installer` instead of the `Vcs`. In later calls it's looking for it as a property on `Vcs`. Can you make the tracker available in both places?

sectioneight:
@crsmithdev would you like me to take over this PR?

crsmithdev:
@sectioneight yes, is there anything I need to do in order to transfer / reassign to you?

sectioneight:
I think you need to add me as a collaborator on https://github.com/crsmithdev/glide so I can push the branch associated with this PR

sectioneight:
@mattfarina added the Tracker onto the VCS as well

sectioneight:
This is not ready for merging yet though, this appears to have broken something:

```
[INFO] Fetching updates for golang.org/x/net.
[INFO] Fetching updates for golang.org/x/sys.
[INFO] Fetching updates for gopkg.in/fsnotify.v1.
[INFO] Fetching updates for gopkg.in/validator.v2.
[INFO] Fetching updates for gopkg.in/yaml.v2.
[INFO] Downloading dependencies. Please wait...
[INFO] Setting references.
[ERROR] Failed to set version on github.com/Sirupsen/logrus to be52937128b38f1d99787bb476c789e2af1147f1: open github.com/Sirupsen/logrus: no such file or directory
[ERROR] Failed to set version on github.com/stretchr/testify to ^1.1.3: open github.com/stretchr/testify: no such file or directory
[ERROR] Failed to set version on github.com/spf13/cobra to : Unable to retrieve checked out version
```

sectioneight:
OK, functionality should be restored now, but I'd like an extra set of eyes on this given that I'm unfamiliar with the code and the path that was broken doesn't seem to be covered by existing tests.

Where would be the best place to add such tests?

mattfarina:
This is now out of sync and about to be superseded by the gps integration work. Closing.

Issue #252, opened 2016-02-12
feature request: `glide up` takes an argument and only updates that package


colemickens:
So, for example, `glide up github.com/Azure/go-autorest` would update the `go-autorest` dependency, without upgrading everything else in the `glide.yaml` file.

mhoglan:
This is similar to what I was referring to in #249 ;  Ability to separate the processing of imports, whether from  `glide.yaml` or the project workspace.

There is an argument that could be made that if you version pinned things in a semantic versioning way, that it shouldn't matter if the latest of that version range was pulled in because it should be compatible.  That is idealistic though, and reality is sometimes you want to just upgrade one dependency.  Especially if your QA process requires verification of every dependency change for a release.

mattfarina:
@colemickens There are two steps to an update in Glide. There are the direct dependencies of the local codebase and then there are the dependencies of the dependencies to form the tree. When we update a direct dependency the dependency tree needs to be re-resolved for any changes. This makes updating just a single dependency difficult. Or rather, it means to make it simple isn't so simple.

We used to have something sorta like this feature. I had not realized it was dropped. We should put something smart back. @technosophos is there a reason i don't know that we dropped this?

atombender:
I'd like this, too.

It's annoying to find that `glide up <some packagename>` runs without complaining about the superfluous argument, but actually updates everything. What is the argument list used for?

tcurdt:
There should be a `glide up <some packagename>` that updates only that dependency - plus the child dependencies.

DallanQ:
I've gotten around this by removing the version number in the lock file for a package that I want to update more frequently than the others. Deleting the package files from the vendor directory followed by `glide install` updates the package to the latest version. This is not a great solution, but I haven't come up with a better way. 

I agree it would really be nice to have `glide up <some packagename>`

atombender:
@mattfarina Any update om this? It's truly annoying to have to undo changes to `glide.lock` when you only want to update a single dependency.

Sometimes I just give up and commit the whole file because I don't know the transitive relationships between packages. For example, if I add a new dependency X, and `glide.lock` now contains 3 new packages, I don't know if these are a result of adding X, or if a coworker had previously forgot to run `glide` when they added a different dependency.

mattfarina:
@atombender it turns out this is more difficult than it seems when you default to versions and ranges. To do this you need to update the one dependency and then walk the tree to make sure the transitive dependencies are picked up and all the versions mesh.

[...]

PR #384, opened 2016-04-17
Initial integration of gps into glide

sdboyer:
This pull request swaps in [gps](https://github.com/sdboyer/gps) for much of the core glide engine.

This checklist is sorta starting in media res of a long-running pull request...but hey, better late than never.

This checklist represents what I think should be necessary to at least merge this PR, though not necessarily be a complete, finished conversion. It might be preferable to merge this into a separate branch, so we can then iterate on other issues in smaller PRs.

- [x] Get Kubernetes working. This is a nice torture test, and so a pretty thorough exercise for gps.
- [x] Store versions alongside revisions in the lock file.
- [x] Add general support (a method? and then, ofc, an underlying field) for reporting the version of a manifest or lock file. (EDIT: there's no explicit version field, b/c that would be a shitshow for users. Instead, we use variations in the format to clearly delineate between legacy and current file formats)
- [x] Create legacy versions of both `cfg.Config` and `cfg.Lock`, and have the unmarshaler detect and report accordingly.
- Fully convert:
  - [ ] `glide install`
  - [ ] `glide get`
  - [ ] `glide update`
- [ ] Remove dead code - the old resolver, etc.


sdboyer:
Yeah, so...the addition of logrus, which in turns pulls in `golang/x/sys` for *one* tiny little thing, is a really gross thing to have to add to a committed vendor dir - it's almost 100k loc. Will try to deal with that later.

sdboyer:
@technosophos @mattfarina feedback on the general approach taken in `Analyzer`, or even just specifically with the godep support I added in c708bf7d75ce8e87295546ced286b8af2b2a2a42, would be great.

Would also help keep the review down to a more manageable size later :)

mattfarina:
@sdboyer first, wow... that's a huge changeset. It's going to take a little bit to look at that.

Second, in your quick review request of c708bf7d75ce8e87295546ced286b8af2b2a2a42, I'll need to look further at why you're asking for the manifest / lock pairing to understand what's going on.

I would prefer that the mechanism that looks for external imports be pluggable rather than hard coded. So, it would iterate over a list of the things rather than go one by one. Though, this is an area the current importer could really improve on.

sdboyer:
@mattfarina the large changset is almost entirely under vendor; as I noted above, it's mostly `golang.org/x/sys`, which is brought in by one tiny point in `Sirupsen/logrus` - and I'm gonna see if I can duck around that by using an earlier version.

If you look at diffs restricted to real, local code additions, there's very little as of now:

```
git diff master --stat -- *.go */*.go
  action/create.go       |   2 +-
  action/ensure.go       |   4 ++--
  action/name.go         |   2 +-
  cfg/config.go          |  60 +++++++++++++++++++++++++++++++++++++++++++++++++++++------
  cfg/config_test.go     |  12 ++++++------
  cfg/lock.go            |  36 ++++++++++++++++++++++++++++++++++++
  dependency/analyzer.go | 105 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
  godep/godep.go         |  46 ++++++++++++++++++++++++++++++++++++++++++++++
  repo/installer.go      |  16 ++++++++--------
  repo/vcs.go            |   2 +-
  10 files changed, 260 insertions(+), 25 deletions(-)
```

> I'll need to look further at why you're asking for the manifest / lock pairing to understand what's going on.

Right, this is the crux of what there is to grok/review. As currently designed, `vsolver` expects pretty much just one thing to be injected by an implementing tool: an instance of `vsolver.ProjectAnalyzer`. That's what I've started building so far, in `dependency/analyzer.go`.

The single method analyzers have, `GetInfo()`, is responsible for returning both manifest-type data, and lock-type data, from what `vsolver` has determined to be the root of a project. (Right now, this is really locked down - repo roots only. I expect that will change, at which point analyzers may also need to provide logic for, "given this path, find the closest project root." I've seen the logic glide has for that, so I'm not worried there - it's just, I'm not quite there yet with the core engine, as supporting it involves more work with the project/package level distinction.

vsolver needs both the manifest and lock as both are relevant to the solving process:

1. [Manifests](https://godoc.org/github.com/sdboyer/vsolver#Manifest) are used from both the root and non-root projects (obviously), as they provide the foundational dep lists and version constraints for which the solver is trying to find a solution.
1. In root projects, manifests also provide three additional things:
  1. Dev dependencies (implemented)
  2. Aliases, which are pretty much what they sound like (not yet implemented)
  3. Overrides, which allow the root package to define global rules that can supercede the constraints given by other projects (not yet implemented)
1. [Locks](https://godoc.org/github.com/sdboyer/vsolver#Lock) are primarily used from the root project, as they provide us the version that we should try to end up with, unless a) something new in the solving process necessitates it change or b) the user explicitly requested an upgrade.
1. However, we also use locks from non-root projects to try to achieve more stability in deeper parts of the depgraph where no root lock has guidance. (not yet implemented)

This last idea is one that, as far as I know, will be unique to glide, as I first suggested in my article. @technosophos and I have discussed it several times as possibly being very helpful in dealing with the go project ecosystem "as-is" - full of locked deps from godep, etc.

> I would prefer that the mechanism that looks for external imports be pluggable rather than hard coded. 

That's probably doable. I don't think it's really either/or, though - we can have the hardcoded analysis for godep, gom, etc., and if those don't work, let a plugin have a go? Or let the plugin have a go first, and then fall back on those. Either way.

TBH though, a use case for this does not immediately spring to mind.

> So, it would iterate over a list of the things rather than go one by one.

I'm confused, "iterating over a list" and "go one by one" seem like synonyms to me. Unless you're suggesting that they all run unconditionally, even if one indicates that it successfully analyzed it?

I guess that's possible. Seems to me you'd have a mess, though, and I'm not clear on what case that's trying to cover. Some project simultaneously using, say, godep and glide, maybe? If anyone is doing that, then erroring out seems like the saner thing to do than trying to make sense of a fundamentally nonsensical situation. (The user could always be encouraged to fork that bizarre project and fix it - which really seems saner to me than trying to have tooling figure out what manner of custom crazy happens from the combination of two systems).

sdboyer:
Just rebased in order to (at least temporarily) get rid of `golang/x/sys` - it's only needed for solaris, so we can deal with it a little later. By rebasing, it's not clogging up the history, and it makes reading the diff more feasible.

mattfarina:
@sdboyer Two pieces of feedback so far...

1. The use of `github.com/Sirupsen/logrus` is going to create an odd UX if debugging is enabled (which it should be). We should talk this part through.
2. Need to deal with dependencies in test files. This is a big ask and someone already started a PR. It's currently the most requested feature.

I'm going to keep poking at this but I wanted to record these before I switched off reviewing for the day.

sdboyer:
> The use of `github.com/Sirupsen/logrus` is going to create an odd UX if debugging is enabled (which it should be). We should talk this part through.

right, so, `logrus` is definitely an issue. ...sort of.

i wrote that logging layer while i was still getting the core solver algorithm right. initially, i imagined that it might be helpful for users, but immediately found (once i started using it to debug the solver) that it was difficult even for ME to use. so, i opened https://github.com/sdboyer/vsolver/issues/5, the outcome of which will specifically be a trace logger, with output looking like the linked file. it will be, i dare say, 🎉 🎉 🎉 

that said, i don't think there's *actually* a problem here. the reason i'm using `logrus` in the first place is because i needed a logging framework that the tool could fully control and inject as a dependency into the solver. while the output right now is mostly useless, both of those things are true - just set the log level to something below `Info`, and you won't hear a peep out of it. this isn't an acceptable long-term solution, of course, but that's why i have that issue open.

> Need to deal with dependencies in test files. This is a big ask and someone already started a PR. It's currently the most requested feature.

we discussed a bit in gitter after you posted this, so reiterating a bit...yes, this is crucial. we already do half of it now - [reading the dev/test deps out of the manifest](https://github.com/sdboyer/vsolver/blob/master/solver.go#L530-L532). the other half is ensuring we incorporate test files in static analysis. the [(untested) code is already (mostly) written](https://github.com/sdboyer/vsolver/blob/master/pkg_analysis.go#L35) for that; i ultimately put it down because the requirement was still a bit murky for me, and i felt that the work required to integrate with glide would clarify it.

so, yeah, the static analysis part is high on the priority list. my plan was to incorporate it if not in this PR, then in the next one.

sdboyer:
Small progress update:

The latest set of changes takes one approach at solving the issue raised in #391. It's provisional, hacky code, but it does the job well enough for now, and it can give some basis for discussion, at least.

In any case, having at least a temporary solution in place for that, I'm now up against the next barrier, which has turned out to be package-vs-project-level analysis. That is, since I'm working against glide itself here, and that annoying `golang.org/x/sys/unix` package is being brought in by `logrus` (as previously discussed)...well, that's not a repo root - `golang.org/x/sys` is.

Tackling this properly means refactoring to include at least some of the static analysis I mentioned in my last comment. There won't be much visible on this side, but there'll be a fair few changes on the `vsolver` side.

mattfarina:
@sdboyer can you use the normal logger interface for your application and pass in logrus for your test runs? Then anything implementing the logger interface can be passed into vsolver. I can even make a logger interface for the `msg` package so it can work fluidly with Glide. I'm happy to discuss the details.

sdboyer:
@mattfarina the logger interface doesn't do leveled logging. I was more convinced that I needed that earlier; now, I'm back to being on the fence. If all we're doing is the trace, leveled logging isn't really necessary. However, there are a *lot* of other classes of errors/information that it may end up being good to emit via logging, rather than just queueing up onto the `Result`, in which case I'll want to stick with leveled logging. 

I won't know what seems right until I start sorting out those errors, though, so It's kinda premature to say right now. For now (read: this week or so 😄 ), I want to stick with a leveled logger.

Fortunately, `logrus` provides plenty of output control options that can be used to make it interoperate nicely with `msg`. Or, as I noted earlier - you can just set the log level down for the moment, and it'll be like it's not there at all.

sdboyer:
Short term solution: 5a71069 turns down the logging level :)

sdboyer:
OK, couple upcoming changes here:

- `logrus` has been removed completely; trace output is done through `*log.Logger`. also, the new trace output is in place.
- `vsolver` now differentiates between a 'local name' (as in, what's literally in import paths) and a 'network name' (where you fetch it from). manifests not agreeing on where a project is sourced from is a failure condition. this should be equivalent to what glide has now.
 - the mechanism to allow a root project to forcibly set the network name and ignore what other projects think isn't in yet, but it will be soon.
 - that override mechanism is different from aliasing - that'll be a little later. in fact, they'll probably both come at the same time.

sdboyer:
soooo...kinda exciting. `glide update` works, at least for a simple case or two :grin:

sdboyer:
We discussed splitting this into separate PRs in order to keep it more manageable. A good goal, but slightly blocked on the state of the deps (semver & vcs). So, in the meantime, here's an overview of what's changed in order to ease review.

First, this is the current status:
```
$ git diff master --stat -- *.go */*.go
 action/create.go       |   2 +-
 action/ensure.go       |   4 +-
 action/install.go      | 110 ++++++++++++++++---------
 action/name.go         |   2 +-
 action/update.go       | 136 +++++++++++++++----------------
 cfg/config.go          | 163 ++++++++++++++++++++++++++++++++-----
 cfg/config_test.go     |  12 +--
 cfg/lock.go            |  41 ++++++++++
 dependency/analyzer.go | 213 +++++++++++++++++++++++++++++++++++++++++++++++++
 gb/gb.go               |  43 +++++++++-
 godep/godep.go         |  46 +++++++++++
 gom/gom.go             |  91 +++++++++++++++++++++
 gpm/gpm.go             |  36 +++++++++
 repo/installer.go      |  16 ++--
 repo/vcs.go            |   2 +-
 15 files changed, 773 insertions(+), 144 deletions(-)
```

In general, changes fall into one of three categories:

- The analyzer (`dependency/analyzer.go`). This is what's injected into the `SourceManager` that's responsible for looking at a tree of files, and extracting manifest and/or lock data from it. All of the changes in `gb`, `godep`, `gom`, and `gpm` serve this purpose.
- Changes under `cfg/*` are all pursuant to making the existing `*cfg.Config` and `*cfg.Lockfile` types fulfill the `vsolver.Manifest` and `vsolver.Lock` interfaces.
- Changes under `action/*` are the refactorings of how `install` and `update`. `install` was working a while ago, and still is afaik; `update` is what I just got working last night.

sdboyer:
Major progress: `update` and `install` are now powered by vsolver!

The basic test I've been running is just against glide itself. Check out this branch, `make build`, then `./glide up github.com/Masterminds/vcs`. The trace logger is currently hardcoded to be on (just temporarily, for dev purposes - this is *NOT* a requirement!), so you'll see the solver work its way through the deps. When it's done, you should have an updated `glide.lock` where only `github.com/Masterminds/vcs` has changed. Boom, #252 et. all, fixed.

So, yay! 🎉 🎉 🎉 🎉 🎉 🎉 

OK, that said...

### Caveats 

There are several. 

1. If you run `glide up` *without* arguments, you'll find your way to a panic. Yup, I'm workin on it - it's a general case that needs a better guard.
2. I'm not dealing with package-level analysis at all yet, though it's pretty much on-deck.
3. I got rid of a number of the params for those two commands, but kept the ones @mattfarina and I talked about. So, they're still there, but I haven't written support for them yet: , 
 * `--cache-gopath`: I just haven't touched this at all yet. Lower priority, though not ignored (#407)
 * `--strip-vendor`: currently, this happens unconditionally. I'm ratcheting that back to something the tool can control: sdboyer/vsolver#24
 * `--all-dependencies`: As above, I've not dealt with package-level analysis yet, but it's coming up, things are mostly in place, and I've got a plan.
 * `--resolve-current`: this is kind of a sticky wicket in general, as @technosophos and I have discussed. I'm planning on punting on this for a while, until the bigger issues are dealt with; it's a whole other dimension of analysis.

To be clear, this is by no means done - though my top priority now is splitting this off into some sub-issues and putting up discrete PRs against a new branch, so this can go more in chunks. However, the path forward is, I think, starting to look pretty clear.

sdboyer:
That last commit fixes the panic described in that first caveat. I'm pretty sure it works, as I used `glide up` to achieve the update that's in that commit :)

sdboyer:
Small status update, since, holy crap it's been a month: some of the delay is that I had a very busy May, but also, bimodal analysis is tricky 😼 . sdboyer/vsolver#36 is where that work is ongoing. i'm decently close - just wrote a new testing harness last night that lets us deal with all the new complexity. while there are some other things to take care of, i think we're on a pretty direct path to integration after that is done.

sdboyer:
last step before merge, i think - simple fixes to static analysis.

sdboyer:
sdboyer/vsolver#36 is now merged...91 freakin commits later. oi.

i'm sure i'll hit some rough spots and bugs when i integrate all that here, but hopefully not anything major. i'll get started in the next few days - i need a bit of a breather right now.

sdboyer:
So, maybe not so much of a breather 😼 

Some notes on that most recent merge:
* The trace logger has been updated a bit, but it really needs another pass, as it's basically still project-oriented in a bimodal world: sdboyer/vsolver#41
* `--strip-vendor` is now something that the tool can control, as promised earlier.
* `--all-dependencies` is...well, we should probably discuss this one. @mattfarina has previously pointed out that this can be handy while developing, as it lets you rapidly pulling in additional packages from existing deps without having to rerun a command. Definitely true, but there's both internal complexity costs and UX costs.
* I've entirely punted on os/arch for right now, but I spent enough time poking at the static analysis system that I know how I want to capture that information. (Still not quite sure how to integrate it in the solving loop, though...)
* We do underlying comparisons on versions through revisions, now. This is most important because it also lets us match revisions - AKA what godep and most other tools encode, and what glide.lock files currently encode - with any equivalent branch or tag identifier. *But it's got a wider scope, too - if you constrain to `v2.0.0`, and someone else constraints to branch `foo`, but those two happen to point to the same revision at the time we solve, then they transparently match. You NEVER get a code version you didn't ask for - it just lessens the surface area for conflicts.*

I'm now catching this up to the latest changes in glide's master.

sdboyer:
Working on adapting `glide get` now. sdboyer/vsolver#42 is a pre-requisite for making that smooth, though we could arguably merge without it and smooth it out later.

sdboyer:
@thockin i'm at the point where testing this against kubernetes would be helpful. have you worked up a glide.yaml for kubernetes, somewhere?

I could convert from Godeps myself, but if you've done the work to have some saner constraints specified, that'd make my testing more realistic and helpful.

sdboyer:
`vsolver` has officially been renamed to `gps` - woohoo!

I've been mostly working on getting kubernetes to work using `gps`. Pretty close, I think, but there are definitely a few more bugs. I'll update this PR again incrementally soon.

sdboyer:
Updated to gps v0.8.1 and resolved conflicts. Back to grinding on getting kubernetes to work...

sdboyer:
Made it all the way through a kubernetes solve last night. Sweeeeet.

mattfarina:
@sdboyer A few things.

1. Keep up the good work.
2. I like the example where someone can use `gps` with just a few (well 30) lines of code.
3. It seems there is still a long way to go on features. For example, the `go-get` redirects. I'm curious of the plan.

If you want we can get on a videochat and talk sometime.

sdboyer:
Thanks! 🎉 🎉 

> It seems there is still a long way to go on features. For example, the `go-get` redirects.

This wasn't implemented for a minute there, but is now, I think. [logic here](https://github.com/sdboyer/gps/blob/master/remote.go#L223-L243), [tests here](https://github.com/sdboyer/gps/blob/master/remote_test.go#L401-L414).

I do plan to refactor that API a bit more - if nothing else that [`remoteRepo`](https://github.com/sdboyer/gps/blob/master/remote.go#L17) type is not how i really want it to work, but i also need to fold it into `SourceManager` - but the logic is, I think, essentially sound.

> I'm curious of the plan.

My plan wrt glide can be seen in the [MVP milestone](https://github.com/sdboyer/gps/milestone/1). If it's in there, then I consider it necessary for glide's completeness (and tool completeness in general). As that list suggests, I don't actually think there's that much more to do.

Big picture, though, this whole thing has gotten quite large, and I haven't necessarily popped out issues for every single issue I've run across. So, if there's a specific feature you see missing, *please please* open an issue for it, or comment on one I already have so that I can see about moving it into that milestone.

Now, gps changes are somewhat separate from actual glide integration itself. Ideally, I'd like to see this merged into a separate branch once the MVP milestone is done (or maybe before). That way we can get out of a monolith and into a more iterative mode when it comes to really baking glide around gps.

sdboyer:
bump - responding to/acting on the following in my last post would be very helpful:

> Ideally, I'd like to see this merged into a separate branch once the MVP milestone is done (or maybe before). That way we can get out of a monolith and into a more iterative mode when it comes to really baking glide around gps.

thoughts on this approach?

> So, if there's a specific feature you see missing, please please open an issue for it.

please?

> This wasn't implemented for a minute there, but is now, I think.

or have i missed something?

ches:
I'm excited for this, I believe it's ultimately the solution for a number of issues with Glide where I just can't predict or have confidence in its current behavior. I don't have any projects nearly of the scope of Kubernetes, but I don't mind testing the branch and trying to give UX feedback or whatever is helpful (including trying to build other public projects).

I believe this is going to change some usage patterns of Glide, by necessity because the solver discussions and implementation have proven some things unsound or irreconcilable. Should there be an outline somewhere of use cases and suggested practices that Glide aims to support, and what breaking changes this may introduce?

sdboyer:
> I don't mind testing the branch and trying to give UX feedback or whatever is helpful (including trying to build other public projects).

:tada: :tada: :+1: :tada: :tada: :tada: :tada: :+1: :+1: :+1: :tada: :+1: :tada: :+1: :tada: :tada: :tada: 

The [last item](https://github.com/sdboyer/gps/pull/83) in my MVP milestone is dragging out a bit, as it's pulled me down into a wider refactor. I'm hoping I can finish by Friday, though. Once I do, I'll return to this, and it would be ENORMOUSLY helpful to have someone else poking, testing on things, and providing feedback.

> I believe this is going to change some usage patterns of Glide, by necessity because the solver discussions and implementation have proven some things unsound or irreconcilable. Should there be an outline somewhere of use cases and suggested practices that Glide aims to support, and what breaking changes this may introduce?

I've been trying to keep notes to this end, but it's a ball I've dropped more than not. Coming up with a list of such things is probably one of my major TODOs after the milestone is done.

sdboyer:
OK, I've closed out the MVP milestone. I didn't finish up that URLs issue, but as I noted on sdboyer/gps#27, reintegrating it here should be a fairly small hiccup.

So, it's time to double-down on getting this all set. @ches, would you be willing to maybe own maintaining that list of differences? we could hop on voice for ~20 minutes to discuss.

sdboyer:
Note that #565 is now open to discuss the major changes that introducing gps into glide will entail.

sdboyer:
The latest set of changes introduce a conversion to the new config and lock file formats. I still have some tweaks to make (upon further, post-merge reflection last night), but it's pretty close.

sdboyer:
The pushed commits should fix those issues. I've run `glide up` a few times now, with no error - hopefully it works the same for you :)

The single biggest issue was that the `glide.yaml` was only halfway converted. Like, it used the new `branch` key for some constraints, but all of the constraints were still under the old key name, `imports`, instead of under `dependencies`. There there was still this line in the output in your gist:
```
[WARN]	glide.yaml was in a legacy format. An attempt will be made to automatically update it.
```

which was an indication that it was using the legacy logic to interpret glide.yaml, causing all `branch` constraint keys to be silently ignored - the legacy handler just drops keys it doesn't know about. Now, while that's a nasty situation, I also think it's not a huge concern for actual users, because a half-converted `glide.yaml` is something that really only seems likely to happen in bootstrappy-type situations - that is, literally only for us, in the process of converting glide to the new system. (Not to say we couldn't/shouldn't necessarily think of ways to improve that autoconversion process - e.g., we could make it fail if the legacy importer sees a `branch` key)

Now, initially I was more concerned by something else: even if the root manifest was halfway converted (and thus it didn't pick up the `branch: 2.x` constraint on `github.com/Masterminds/semver`), gps has _always_ relied on that version of semver, and included it in its manifest since the beginning. So, why wasn't glide picking up that constraint when visiting gps' manifest and applying it to the depgraph, causing us to get the right version of semver anyway?

Turns out, the answer's in the trace:

```
| | | | | | ? attempt github.com/sdboyer/gps with 1 pkgs; 30 versions to try
| | | | | | | try github.com/sdboyer/gps@v0.10.1
| | | | | | | ✗ github.com/sdboyer/gps@v0.10.1 depends on github.com/Masterminds/semver with 2.x, but that's already selected at v1.1.1
| | | | | | | try github.com/sdboyer/gps@v0.10.0
| | | | | | ✓ select github.com/sdboyer/gps@v0.10.0 w/1 pkgs
```

The halfway converted root `glide.yaml` specified `branch: master` for gps, but for the same reasons above with semver, that constraint got silently dropped. With no explicit constraint, we instead accept anything, which means the solver started working from the most recent semver downwards, beginning with `v0.10.1`. And indeed, the `glide.yaml` in gps@v0.10.1 is fully converted and correct, declaring that constraint on `semver` to be 2.x. But, because we'd already selected `v1.1.1` earlier on, that was a conflict, so the solver did what the solver does: tried the next version of gps, `v0.10.0`, to see if that might work.

And that's where things got fun, because `v0.10.0` ALSO had a halfway-converted manifest:

```yaml
package: github.com/sdboyer/gps
owners:
- name: Sam Boyer
  email: tech@samboyer.org
import: # legacy name
- package: github.com/Masterminds/semver
  branch: 2.x # new key
  vcs: git
- package: github.com/Masterminds/vcs
  vcs: git
- package: github.com/termie/go-shutil
  version: bcacb06fecaeec8dc42af03c87c6949f4a05c74c
  vcs: git
```

So, we had the same problem repeating itself a layer down - glide's analyzer autoconverted that glide.yaml, silently dropped the branch constraint, which made it look like v0.10.0 was cool with any version of semver, which of course meant that there was no longer a conflict with the already-selected semver@v1.1.1 - so the algo picked gps@v0.10.0, and off we went.

Again, I think this is mostly a problem that's unique to the bootstrapping situation. And also, just to explain, the reason these half-conversions existed at all is because, to deal with the ambiguity of semver-or-branch discussed in #391 specifically wrt `github.com/Masterminds/semver`, I introduced logic to allow that `branch` key _long_ prior to formally creating the new/legacy format distinction. So, when those manifests were written, they actually did work. They just don't now, because glide's logic has changed.

However, this particular incident does change some of my feelings on how glide's analyzer is currently written. Right now, it tries really hard _not_ to error on `DeriveManifestAndLock()`, because my thought was that it's better to include versions that might be right than to exclude them. This experience, however, even if it is mostly a product of bootstrapping, demonstrates just how fast those problems can become harmful and impossible for an end-user to debug. So, I think it's probably better to be conservative in those cases - if the metadata looks _at all_ bad, just throw the version out.

mattfarina:
This has been merged into the [gps-integration](https://github.com/Masterminds/glide/tree/gps-integration) branch. We can continue the development there. Smaller PRs and incremental work.

PR #390, opened 2016-04-20
Added support for custom Go executable name.

dpmcnevin:
The Google App Engine SDK ships with a python wrapper called goapp instead of the regular go executable.

Usage example:
`GLIDE_GO_EXECUTABLE=goapp glide install`

Current workaround is to install both go and go-app-engine-64 packages.


mattfarina:
@dpmcnevin thanks for the contribution

PR #392, opened 2016-04-22
Correctly normalize Windows package paths.

### LINKED ISSUE #389: glide update writes subpackages in lockfile differently on Windows than Unix
On Unix, subpackages are written like this:

```
-name github.com/foo/bar
 version: xxxxx
 subpackages:
 - baz
 - baz/qux
```

However, on Windows, the subpackages are written as:

```
-name github.com/foo/bar
 version: xxxxx
 subpackages:
 - github.com\foo\bar\baz
 - github.com\foo\bar\baz\qux
```

Notice how the full package path is used with backslashes, even though the repo name uses forward slashes.

If an update is performed on Windows on a lockfile created on a Unix machine, the subpackages are duplicated:

```
-name github.com/foo/bar
 version: xxxxx
 subpackages:
 - baz
 - baz/quz
 - github.com\foo\bar\baz
 - github.com\foo\bar\baz\qux
```

Again, notice how the original Unix subpackages are using forward slashes while the ones written on the Windows host are using backwards slashes.

I haven't noticed any issues with glide resulting from these differences (yet), but as I develop on both operating systems I would prefer the output to be the same from each.

This was experienced using glide master (as of 0f537107d21b9e5854039b95ec9f239f286e5d89).

jrick:
When the name passed to util.NormalizeName contains backslashes from a
Windows filepath, the backslashes must be first converted to forward
slashes (used by all package names, including on Windows).  This fixes
the strings.TrimPrefix call used to strip the repo root from the
subpackage.

As a result of this, subpackages in lockfiles are written identically
on Windows as other Unix platforms and no duplicate subpackages are
introduced.

Fixes #389.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -311,6 +311,7 @@ func NormalizeName(name string) (string, string) {
 		}
 	}
 
+	name = filepath.ToSlash(name)
 	root := GetRootFromPackage(name)
 	extra := strings.TrimPrefix(name, root)
 	if len(extra) > 0 && extra != "/" {
--- a/util/util.go
+++ b/util/util.go
@@ -30,11 +30,7 @@ func init() {
 }
 
 func toSlash(v string) string {
-	v = strings.Replace(v, "\\", "/", -1)
-	if os.PathSeparator != '\\' && os.PathSeparator != '/' {
-		v = strings.Replace(v, string(os.PathSeparator), "/", -1)
-	}
-	return v
+	return strings.Replace(v, "\\", "/", -1)
 }
 
 // GetRootFromPackage retrives the top level package from a name.
--- a/util/util.go
+++ b/util/util.go
@@ -29,13 +29,21 @@ func init() {
 	}
 }
 
+func toSlash(v string) string {
+	v = strings.Replace(v, "\\", "/", -1)
+	if os.PathSeparator != '\\' && os.PathSeparator != '/' {
+		v = strings.Replace(v, string(os.PathSeparator), "/", -1)
+	}
+	return v
+}
+
 // GetRootFromPackage retrives the top level package from a name.
 //
 // From a package name find the root repo. For example,
 // the package github.com/Masterminds/cookoo/io has a root repo
 // at github.com/Masterminds/cookoo
 func GetRootFromPackage(pkg string) string {
-	pkg = filepath.ToSlash(pkg)
+	pkg = toSlash(pkg)
 	for _, v := range vcsList {
 		m := v.regex.FindStringSubmatch(pkg)
 		if m == nil {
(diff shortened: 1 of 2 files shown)


util/util.go
```diff
@@ -311,6 +311,7 @@ func NormalizeName(name string) (string, string) {
 		}
 	}
 
+	name = filepath.ToSlash(name)
 	root := GetRootFromPackage(name)
```
mattfarina: What do we get from this? The first thing `GetRootFromPackage` does is run `filepath.ToSlash`. `GetRootFromPackage` is sometimes called outside `NormalizeName`.

Is there a problem in the latest code that needs to be addressed?

jrick: The issue is that name contains backslashes.

With `github.com\foo\bar\baz` as input:

```
// name == "github.com\foo\bar\baz"
root := GetRootFromPackage(name) // root == "github.com/foo/bar"
extra := strings.TrimPrefix(name, root) // extra == "github.com\foo\bar\baz"
```

jrick:
Oops, intended to add a testcase for this and it fails because it just concats the root + extra, which no longer matches the initial input.  Fixing shortly.

util/util.go
mattfarina: I see the bug *I* introduced now.

mattfarina:
So, this is entirely a micro-optimization but...

Since the root name is already being normalized then the extra is the only part that needs to have it done. Can that happen [here](https://github.com/Masterminds/glide/blob/cdc7a9e51ffe7bf45f3efa1e1ab1b2f2d3c82ae7/util/util.go#L317) instead?

jrick:
No. That will convert the backslashes to forward slashes, but the returned subpackage will still contain the full root since the TrimPrefix just above it did not remove the root.

jrick:
I'm not sure why it's failing on Travis now.  Passes locally.

jrick:
Oh.. I see.

From filepath/path.go:

```
func ToSlash(path string) string {
    if Separator == '/' {
        return path
    }
    return strings.Replace(path, string(Separator), "/", -1)
}
```

`Separator` is `os.PathSeparator` so this has different behavior on Windows vs Unix.  Specifically, on Unix, it won't replace backslashes with with forward slashes since backslashes are not a separator character on those platforms, causing the tests to fail there.

mattfarina:
Maybe we need to tweak the use of [`normalizeSlash`](https://github.com/Masterminds/glide/blob/815d933fc6de40fb433c05b5e717716f0129013d/cfg/config.go#L548).

jrick:
Looks like we could just use that instead of `filepath.ToSlash`.

util/util.go
```diff
@@ -29,13 +29,21 @@ func init() {
 	}
 }
 
+func toSlash(v string) string {
+	v = strings.Replace(v, "\\", "/", -1)
+	if os.PathSeparator != '\\' && os.PathSeparator != '/' {
+		v = strings.Replace(v, string(os.PathSeparator), "/", -1)
+	}
```
mattfarina: Is there an operating system with the path separator other than `/` or `\` that Go supports? Or that's in wide use? If so that's a separate issue because other things will fail.

jrick: Not to my knowledge.  Should this only replace backslashes then?

mattfarina: I think using `v = strings.Replace(v, "\\", "/", -1)` should be enough.

mattfarina:
@jrick thanks for the contribution!

PR #394, opened 2016-04-25
fix typo

shawnps:


mattfarina:
@shawnps thanks for the contribution.

Note, the failed test was due to a bug in Glide that was picked up on Go tip. It was fixed in #395.

PR #416, opened 2016-05-11
Remove incorrect info from README.

technosophos:


technosophos:
@mattfarina feel free to merge this whenever you have a chance to review it. I removed an entire section of the README that looks like it was from the pre-Go 1.5 days

PR #417, opened 2016-05-12
Adds golang.org to vcsList

zjx20:
@mattfarina I have trouble to access golang.org in my network environment, so I use aliasing for it.

For example:
```yaml
- package: golang.org/x/net
  repo: https://github.com/golang/net.git
```

However, it doesn't work well with subpackages:
```
[WARN] golang.org/x/net/context appears to be a vendored package. Unable to update. Consider the '--update-vendored' flag.
[WARN] Unable to set version on golang.org/x/net/context to . Err: Cannot detect VCS
```

`glide` didn't realize that the root package of `golang.org/x/net/context` was `golang.org/x/net` which has been downloaded.

Explicitly adding golang.org into vcsList can fix this problem. I would be appreciated if you could accept this PR.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -215,6 +215,10 @@ var vcsList = []*vcsInfo{
 		pattern: `^(?P<rootpkg>hub\.jazz\.net/git/[a-z0-9]+/[A-Za-z0-9_.\-]+)(/[A-Za-z0-9_.\-]+)*$`,
 	},
 	{
+		host:    "golang.org",
+		pattern: `^(?P<rootpkg>golang\.org/x/[A-Za-z0-9_.\-]+)(/?[A-Za-z0-9_.\-]+)*$`,
+	},
+	{
 		host:    "go.googlesource.com",
 		pattern: `^(?P<rootpkg>go\.googlesource\.com/[A-Za-z0-9_.\-]+/?)$`,
 	},
--- a/util/util.go
+++ b/util/util.go
@@ -216,7 +216,7 @@ var vcsList = []*vcsInfo{
 	},
 	{
 		host:    "golang.org",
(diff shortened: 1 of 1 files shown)


mattfarina:
@zjx20 What is the issue you're seeing? `golang.org/x/net` uses `go-get` redirects. That does require network requests to resolve.

There may be a place we can cache some data to avoid lookups.

zjx20:
@mattfarina As you may know the GFW, a firewall at the network backbone in China. Unfortunately, golang.org is blocked by the firewall. We just can't access to golang.org:
```shell
$ curl https://golang.org/x/net/context?go-get=1
curl: (7) Failed to connect to golang.org port 443: Operation timed out
```

So calling `getRootFromGoGet("golang.org/x/net/context")` will just return `"golang.org/x/net/context"` in my computer, but `"golang.org/x/net"` is expected.

I know this PR is too specific, I totally understand if you don't accept it. But if you do, I will be appreciated :)

zjx20:
@mattfarina I can use `glide mirror` to solve the issue now. Thank you very much!

gloomyzerg:
@zjx20  How to configure glide mirror? to break GFW?  thanks !

zjx20:
@gloomyzerg I'm afraid glide mirror isn't the silver-bullet for GFW, I had that illusion just because GFW temporarily unblock golang.org at that time (as of 7 Sep 2016).

Since this PR hasn't been accepted, you can use my own fork of glide (https://github.com/zjx20/glide), and then use the repo aliasing to break the GFW.

Alternatively, if you have a socks5 proxy or a ssh tunnel, you can use this tool (https://github.com/zjx20/socks-git) to solve the problem out of the box.

gloomyzerg:
@zjx20  Thank you very much!

PR #420, opened 2016-05-12
Fixed #396: Don't update the lock file if nothing has changed

### LINKED ISSUE #396: Don't change the lock's timestamp if nothing else changed
Per the discussion in #366, if nothing else has changed when it comes time to write out a lock, then the timestamp shouldn't change either.

mattfarina:


cfg/lock.go
```diff
@@ -45,9 +47,46 @@ func (lf *Lockfile) WriteFile(lockpath string) error {
...
+	n.Updated = lf.Updated
+	n.Imports = lf.Imports.Clone()
+	n.DevImports = lf.DevImports.Clone()
+
+	return n
+}
+
+// Fingerprint returns a hash of the contents minus the date. This allows for
+// two lockfiles to be compared irrespective of their updated times.
+func (lf *Lockfile) Fingerprint() (string, error) {
```
sdboyer: sorta-nit: I wonder if returning `[]byte` is more idiomatic than returning `string` here?

I realize this is consistent with the similar method on configs, but as long as a new instance of it is being added, I figured I'd raise the question.

mattfarina: @sdboyer can you point me to idiomatic examples of fingerprints/hashes as `[]byte`? I'd like to read them.

mattfarina: @sdboyer I did some digging and I see what you mean. But, it raised a concern for me.

What would be returned is a `[32]byte`. To compare the byte slices you could use something like `bytes.Equal`. But, that needs a type of `[]byte`. So to handle a comparison you'd end up with...

```go
bytes.Equal(a[:], b[:])
```

The `a[:]` is shorthand for `a[0:len(a)]`. This is needed to convert the type from `[32]byte` to `[]byte`. For anyone using this function this seems non-intuitive.

Do we know why the standard library does things this way?

jrick: If you use `sha256.Sum256` instead of using the `hash.Hasher` interface, you can return a `[32]byte` or `*[32]byte` instead of a slice.  Arrays of equal size can be compared with `==` so there's no need to use `bytes.Equal` for them.

Issue #408, opened 2016-05-06
Any suggestion about installing glide in travic-ci?


yanyiwu:
glide is so good, but its installation is so hard.

yanyiwu:
can you simplify the installation process of glide?

krancour:
fwiw, have a look at github.com/deis/docker-go-dev.  It's a fully loaded, containerized go development environment.  Then look at any project that uses it (e.g. github.com/deis/router).  We use the Travis build type `generic`, require Docker, and use carefully crafted Makefiles that delegate the build to the container.  It solves the problem of how to elegantly install glide on Travis because you will no longer have to... but you also realize many other benefits-- namely your contributors no longer need to jump through hoops to maintain a _local_ go dev environment that's compatible with your project.  This is by no means the "official" answer to this question, but you may find it helpful nonetheless.

jrick:
glide itself is `go get`able (with vendored deps checked into git), it just won't include the exact commit hash in the binary this way.  If you're ok with always pulling the glide master branch, this should be sufficient.

yanyiwu:
@jrick Thank you sooooo much!!!
 `glide itself is go getable` should be written in README.md, because it is so useful for any newbee of glide

krancour:
> If you're ok with always pulling the glide master branch, this should be sufficient.

This is something to be cautious of.  You can't guarantee your tests and builds are deterministic if the version of your dependency manager can change underneath you from one run to the next.

davecgh:
@krancour:  That a good point.  It can be solved by checking out a specific revision after you use `go get` to obtain glide.  For example, in `.travis.yml` you could do:

```YAML
install:
  - go get -v github.com/Masterminds/glide
  - cd $GOPATH/src/github.com/Masterminds/glide && git checkout e73500c735917e39a8b782e0632418ab70250341 && cd -
  - glide install
...
```

Then, whenever you want to update your build environment to a new version of glide, you would simply need to update the `.travis.yml` to newer commit hash.

mattfarina:
An alternative if you want a specific version is...
```yaml
install:
  - wget "https://github.com/Masterminds/glide/releases/download/0.10.2/glide-0.10.2-linux-amd64.tar.gz"
  - mkdir -p $HOME/bin
  - tar -vxz -C $HOME/bin --strip=1 -f glide-0.10.2-linux-amd64.tar.gz
  - export PATH="$HOME/bin:$PATH"
```

willis7:
[...]

PR #428, opened 2016-05-18
Update README.md

yanyiwu:


yanyiwu:
https://github.com/Masterminds/glide/issues/408

mattfarina:
@yanyiwu the problem with this is it's telling people to install the tip of master rather than a [release version](https://en.wikipedia.org/wiki/Software_versioning). We don't ask people to install Go at the tip of master so why should we ask people to install Glide that way?

yanyiwu:
To the most users, they meet this project at first time, they want to find the way to use it as easier as possible , when they can not find out, they will leave.

when i introduce this project to others, they alway complain installation is too sophisiticated. 

when we use codes in our project ,the version of codes is important, when we use a tool, for example,vim,brew or others, we dont care about their version as long as it can work.

sdboyer:
> when we use a tool, for example,vim,brew or others, we dont care about their version as long as it can work.

Sorry, but this is a false comparison.

You might not care about what version of vim, brew, or others that you're using, but that's because you ARE getting stable versions of them - which is a guarantee provided by the design of the system you're using to get those packages. `

`go get` is *not* designed to get stable versions - just, whatever the latest `HEAD` is. Ironically, this is one of the arguments for having something like glide exist in the first place.

The best, most stable thing that could probably be done is providing a script that allows installation via something like `curl <URL> | sh`.

mattfarina:
@yanyiwu You bring up a good point. Making it easier to install would be useful.

When you install brew, etc you get the latest release rather than the tip of development. That's my concern.

Today you can:
* Install via a binary
* Install with `brew install glide`
* Add an apt repo and install with `apt-get install glide` under ubuntu.

What do you think of the method brew and [helm](http://helm.sh/) provide with a script and a native windows method?

franciscocpg:
I think that @sdboyer suggestion (`curl <URL> | sh`) is really a good option.
But for this to work we need to have releases for all supported platforms.
According to this [link](https://golang.org/src/go/build/build.go?s=#L260), here is the list of supported (cgo) platforms:
```
"darwin/386":      true,
"darwin/amd64":    true,
"darwin/arm":      true,
"darwin/arm64":    true,
"dragonfly/amd64": true,
"freebsd/386":     true,
"freebsd/amd64":   true,
"freebsd/arm":     true,
"linux/386":       true,
"linux/amd64":     true,
"linux/arm":       true,
"linux/arm64":     true,
"linux/ppc64le":   true,
"android/386":     true,
"android/amd64":   true,
"android/arm":     true,
"netbsd/386":      true,
"netbsd/amd64":    true,
"netbsd/arm":      true,
"openbsd/386":     true,
"openbsd/amd64":   true,
"solaris/amd64":   true,
"windows/386":     true,
"windows/amd64":   true,
```
@mattfarina: By the way, how are you doing the releases?

mattfarina:
@franciscocpg releases are generated by running `make dist`. See https://github.com/Masterminds/glide/blob/master/Makefile#L30. It should be fairly easy to build distributions for other environments and create a shell script (for those crazy enough to run them).

franciscocpg:
Ok @mattfarina 
I am going to start working on this.

mattfarina:
I don't really want to advocate installing the latest dev snapshot as the primary method.

There is now a shell script that can install on *nix. It's in homebrew and we have a PPA.

Next steps are for binaries for more platforms (freebsd, arm, etc). There are other issues for that.

Would also be good to have a oneget install for Windows (e.g., `Install-Package glide`). But, I think that's for a future issue.

Issue #430, opened 2016-05-19
Proposal: Drop subpackages from glide.yaml file


mattfarina:
This idea is to remove the record from the `glide.yaml` file while keeping the record in the `glide.lock` file.

The `subpackages` from the `glide.yaml` used to be used as part of the `rebuild` command for the setup prior to the `GO15VENDOREXPERIMENT`. That command is no longer needed (instead `go install` should be used).

Thoughts @technosophos?

xeger:
The problem with `go install` is that I can't version a code-gen tool binary (`goagen`) alongside a vendored library dependency (`goa`). Makes things a bit difficult to work with. Is there a best practice for this + glide?

elan100cs:
One possible use case of specifying subpackage is when we want to force to pull the dependency of that subpackage. This is useful when a project define 1 repository for all of it's common library, which doesnt guarantee for all of the subpackages gets imported within the root packages.

Since now, the way to use glide is to specify it's root package, i find that if we dont have any imports of the subpackage within the root package, the dependencies of the subpackage doesnt gets fetch automatically, though i thought the old version of glide does this automatically.

Can we use subpackage to fulfill this use case? Or does it already being handled by some other features in glide ?

PR #431, opened 2016-05-19
Updating the docs on subpackages

mattfarina:


docs/glide.yaml.md
```diff
@@ -41,6 +41,7 @@ These elements are:
     - `version`: A semantic version, semantic version range, branch, tag, or commit id to use. For more information see the [versioning documentation](versions.md).
     - `repo`: If the package name isn't the repo location or this is a private repository it can go here. The package will be checked out from the repo and put where the package name specifies. This allows using forks.
     - `vcs`: A VCS to use such as git, hg, bzr, or svn. This is only needed when the type cannot be detected from the name. For example, a repo ending in .git or on GitHub can be detected to be Git. For a repo on Bitbucket we can contact the API to discover the type.
+    - `subpackages`: A record of packages being used within a repository. This does not include all packages within a repository but rather those being used.
```
silasdavis: Could we have a note to say why we might care. As I understand it it is prevent all the (transitive) dependencies being fetched for a package within a repo that you don't use. Is that all?

mattfarina: @silasdavis There is not as much of a reason to have it as there used to be. This was far more important prior to the code scanning we have now.

This is useful in the lock files because other tools can use that information such as glide-vc.

xeger:
@mattfarina I left a comment on #430 about the future of subpackages; I understand that their original purpose has fallen by the wayside, but there's an itch that's left unscratched, relating to versioning the binaries that one `go install`s by hand, and their dependencies on vendored source... Not sure that it's glide's place to scratch this itch but I was hoping to hear your perspective.

PR #433, opened 2016-05-20
A managed global Glide cache

mattfarina:
This is the first step. Follow-on PRs will take advantage of this cache with more speed and to have a better UX.


mattfarina:
@technosophos if you have 10 minutes to review.

sdboyer:
this seems like a weird thing to work on, if vsolver integration is indeed coming soon?

PR #437, opened 2016-05-22
extract guessImportDeps function

klnusbaum:
`guessDeps` is a rather long function (and there is a comment at the top acknowledging so). Let's try to extract to some of the functionality out of it to make it easier to read. This groups all of the Import related dependency guessing into a single function called `guestImportDeps` and moves that logic out of the `guessDeps` functions, making things more concise in `guessDeps`.

My long term plan is to keep refactoring this file until it's easier to test. This is a nice step in the direction of making this file more testable.


mattfarina:
@klnusbaum thanks for the contribution.

PR #438, opened 2016-05-22
add ability to read symlink basedirs

klnusbaum:
Imagine the following scenario: A user has a Go project. This directory is sitting in an arbitrary location in their file system and the user would like to keep it there. However, they need it to work with Go tooling, so they symlink it into an appropriate location in their GOPATH, e.g. `$GOPATH/src/github.com/user1/myproject`. The user then decides they's like to start using glide, so they cd to `$GOPATH/src/github.com/user1/myproject` and run `glide create` to generate their initial `glide.yaml` file. The result is the creation of empty `glide.yaml` file. Sad times.

What's happening is that the `basedir` given to `NewResolver` is actually a symlink. This in turn causes an issue in the `ResolveLocal` function. The `if !fi.IsDir()` check fails since the basedir isn't actually a directory, it's a symlink.

Let's add a check to the `NewResolver` function. After getting the absolute path of the provided `basedir`, let's check to see if it's a symlink. If it is, read the symlink and use that as the basedir instead. If not, proceed as normal.

### DIFF
--- a/dependency/resolver.go
+++ b/dependency/resolver.go
@@ -166,6 +166,13 @@ func NewResolver(basedir string) (*Resolver, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	basedir, err = checkForBasedirSymlink(basedir)
+
+	if err != nil {
+		return nil, err
+	}
+
 	vdir := filepath.Join(basedir, "vendor")
 
 	buildContext, err := util.GetBuildContext()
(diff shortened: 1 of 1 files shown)


mattfarina:
I really like this change.

Unfortunately, I case where project can be added to itself. For example,

``yaml
package: github.com/foo/bar
import:
- package: github.com/foo/bar
  subpackages:
  - baz
  - qux
```

I discovered this on a private project so I can't share.

Once we get this fixed I'll be happy to merge this. I'll try to dig into it more later this week. If you find the solution before then that would be great, too.

klnusbaum:
Hey @mattfarina, any updates on this?

mattfarina:
@klnusbaum I've been traveling and then had a holiday weekend so I got a little behind. Need to figure out that bug so we can merge this.

mattfarina:
@klnusbaum I found the problem, fixed it, and merged into master. Thanks for the contribution.

klnusbaum:
awsome! Thanks @mattfarina !

PR #444, opened 2016-05-25
Rework create/init

mattfarina:


mattfarina:
@thockin @technosophos I'd be really curious to know what you think about this? It's a big change designed to be helpful when setting up a project to use Glide.

thockin:
Can you summarize the change?  I'm not equipped for this as a code review.

mattfarina:
@thockin Rather than a code review I'd be curious as a user usage review. So, grab the code, install it, and try `glide init` on a project. Particularly one that imports from Godep.

thockin:
Can you summarize?  I'm beyond buried at the moment, and coming up on a
vacation.  If you want feedback before July...

mattfarina:
@thockin if you're buried and coming up on a vacation don't feel pressure to review.

This turns `glide init` into a wizard. It will try to detect versions from commit ids, try to offer release versions if nothing is detected, and help you setup using ranges for projects that use releases. It's a start and i'm curious on the wizard interaction model.

And, it's interactive. For example, if a project uses semver your choice for ranges may be to follow patch or minor releases.

thockin:
That's interesting!  I think I would want first to do a transliteration -
don't change ANYTHING.  Let me convince myself it compiles and works.
Commit that.  Then I might want to optimize more (e.g. using tags instead
of hashes, or ranges)

mattfarina:
@thockin That alone is fantastic feedback. This is the kind of thing I was looking for.

thockin:
Always ask "what would a paranoid person who has been repeatedly burned by
trusting ANYONE want?".  That's me. :)

sdboyer:
> Always ask "what would a paranoid person who has been repeatedly burned by
trusting ANYONE want?".  That's me. :)

this calls to mind a reddit conversation we had :)

mattfarina:
@technosophos and anyone else... I would love feedback on this new version. Less intrusive and you can use the wizard later.

technosophos:
:ship: I like it.

PR #451, opened 2016-05-31
export GO15VENDOREXPERIMENT only on 1.5

tugberkugurlu:
I am not entirely sure if this is the correct explanation in the context of this project but https://github.com/golang/go/wiki/PackageManagementTools suggests that `GO15VENDOREXPERIMENT` is for 1.5 Go release. 1.6 release makes the vendoring the default behavior.


mattfarina:
@tugberkugurlu In go 1.5 you needed to opt-in with the environment variable, in go 1.6 you can opt-out by setting it to 0, and in go 1.7 it will be on all the time no matter what.

PR #453, opened 2016-06-01
Fix DeleteUnused flag

s-urbaniak:
Hi,

We are evaluating moving over to glide for the rkt project. While testing bits out I found that the `--delete` flag in the `glide update`/`glide install` subcommand actually didn't do anything.


mattfarina:
@s-urbaniak thanks for fixing the bug.

PR #457, opened 2016-06-04
suggestion about the README.md

mh-cbon:
suggestion about the manual install procedure.

Not sure what this is: -X main.version=0.10.2-86-g5865b8e, its relevance, importance, etc


mh-cbon:
I noticed that it works wih go 1.6.2, but not 1.6. not sure why..

Issue #448, opened 2016-05-28
Binary for Linux ARM


wiget:
Please, add binary release for Linux ARM.

mattfarina:
There are multiple ARM architectures. Are you asking for arm64 or a different variant?

wiget:
Suitable for my BananaPi - armv7h :smile: . Users of Raspberry Pi 2/3 will benefit from this too.

mattfarina:
Do you mean [`GOARM=7`](https://github.com/golang/go/wiki/GoArm)?

wiget:
Yes

cescoferraro:
armv6 is the rpi2.
armv7 is for the rpi3

mattfarina:
I've been using [gox](https://github.com/mitchellh/gox) for cross compilation. It seems gox doesn't let you set the GOARM variable. Will need to look into alternative setups to build the release.

cescoferraro:
@mattfarina have you filed it as a bug upstream? it seems odd

franciscocpg:
Hi @cescoferraro 
I opened this [issue](https://github.com/mitchellh/gox/issues/67) proposing a PR, but no response yet.

mattfarina:
With the v0.12.0 release there are arm binaries.

Issue #445, opened 2016-05-26
Repo for Raspbian Wheezy arm??


cescoferraro:
```
W: Failed to fetch http://ppa.launchpad.net/masterminds/glide/ubuntu/dists/wheezy/main/source/Sources  404  Not Found

W: Failed to fetch http://ppa.launchpad.net/masterminds/glide/ubuntu/dists/wheezy/main/binary-armhf/Packages  404  Not Found

E: Some index files failed to download. They have been ignored, or old ones used instead.
```

franciscocpg:
Hi @cescoferraro 
As mentioned in the README of this project there is repo for these ubuntu distros (and here is the project https://github.com/Masterminds/glide-ppa)
`Ubuntu Precise(12.04), Trusty (14.04), Wily (15.10) or Xenial (16.04) `
Are you asking for a debian wheezy repo?

mattfarina:
@cescoferraro Are you asking for arm64 or some other architecture? [Go has numerous arm options](https://github.com/golang/go/wiki/GoArm).

@franciscocpg thanks for jumping on this.

cescoferraro:
@franciscocpg yeah, wheezy is gonne. Jessie would be better. I am buliding it with go for now.

franciscocpg:
I think this issue is related to issue https://github.com/Masterminds/glide/issues/448 and PR https://github.com/Masterminds/glide/pull/428.
I see no problem adding the wheezy dist to ppa, but I really think that `curl <URL> | sh` script with the github release is a better option to reach all supported golang platforms.

cescoferraro:
@mattfarina rpi 2 uses armv6 while the rpi3 uses armv7.

mattfarina:
When I do the next release I'll try to make the installation process better.

mattfarina:
With the v0.12.0 release there are arm binaries. Does this work (you should be able to use the install script).

PeterBurner:
Install script is not working for armv7
Using Raspberry PI 3 Model B+ with Raspbian  2019-07-10
```
ARCH=armv7
OS=linux
Using curl as download tool
Getting https://glide.sh/version
TAG=v0.13.3
GLIDE_DIST=glide-v0.13.3-linux-armv7.tar.gz
Downloading https://github.com/Masterminds/glide/releases/download/v0.13.3/glide-v0.13.3-linux-armv7.tar.gz
Did not find a release for your system: linux armv7
Trying to find a release on the github api.
[...]

PR #467, opened 2016-06-09
Curl install

franciscocpg:
As mentioned at https://github.com/Masterminds/glide/issues/445, https://github.com/Masterminds/glide/issues/448 and https://github.com/Masterminds/glide/pull/428 this PR is about doing a shell script to download and install the latest glide release for a given platform (OS + ARCH).
I tested this script with linux, Mac, FreeBSD and my raspberry pi (linux/arm) successfully.
I made a fake release (https://github.com/franciscocpg/glide/releases/tag/0.10.3) for this testing.


mattfarina:
@franciscocpg thanks for putting this together. I want to give this a good once over and I can put the shell script on `glide.sh`.

install.sh
```diff
@@ -0,0 +1,68 @@
...
+	fi
+	exit $result
+}
+
+# Execution
+
+#Stop execution on any error
+set -e
+trap "bye" EXIT
+testRoot
```
mattfarina: Why does one need to be root?

mattfarina:
I like where this is going but I'd like a few changes...

1. the version URL should be `https://glide.sh/VERSION`. That could be a text file with just the tag in it. A simple PR to that repo would make that work. That would let us control the release.
2. The install script be either in the glide site repo or in a separate repo. I think I prefer the first option. Then it can be coupled to the version script.
3. It should be executed as non-root but ask for password if someone wants to install (and ask for location) and root is needed to do so. This last part could be a separate flow after just getting it works.

franciscocpg:
Hi @mattfarina.
One need to be root because i am forcing location to `/usr/local/bin`.
If we let user decide where to install, then he can decide if it needs to execute as sudo or not.
But I still think it's a good practise for *nix users to install things at `/usr/local/bin`.

Or we can take an approach like rvm (https://rvm.io/rvm/install and https://get.rvm.io).
If the user is not running the script as root then they call `Single-User Install` and install location is `~/.rvm`. If the user is running as root then they call  `Multi-User Install` and the install location is `/usr/local/rvm`.
Another thing to take note is that https://get.rvm.io just redirects to https://raw.githubusercontent.com/wayneeseguin/rvm/master/binscripts/rvm-installer.

mattfarina:
@franciscocpg Thanks for the rvm pointer. Had not looked at that one, yet.

[Helm](http://helm.sh), [Brew](http://brew.sh/), and rvm all seem to do this a little differently.

I'm thinking out loud but I envision...
1. It downloads the right version of glide and puts the binary in the current working directory (telling you that)
2. prompts for the place to put it (defaults to /usr/local/bin).
3. Prompts for password if needed to copy there.

Or, should it put it in `$GOPATH/bin`?

If it fails to copy to the final location it lets you know that it couldn't and tells you where it's at.

Thoughts? I'm always wary magic or asking for passwords on internet shell scripts.

franciscocpg:
@mattfarina 
To be honest I loved your idea about `$GOPATH/bin`.

Golang is a pre-requisite for using glide so its make sense.

Besides that, putting in `$GOPATH/bin` reduces the complexity because we only need to verify if `$GOPATH` env is set.
So, if `$GOPATH` is not set then we fail with a message otherwise we install at  `$GOPATH/bin`.

What do you think?

mattfarina:
@franciscocpg sounds good to me. Want to make the request on the project at https://github.com/Masterminds/glide.sh

Also, the part about doing the builds... it doesn't appear that gox properly handles arm builds. For arm there are different architectures and those don't appear to be happening. Any thoughts on that?

franciscocpg:
@mattfarina 
According to this link https://github.com/golang/go/wiki/GoArm#supported-architectures, we need only to set  `GOARM` variable.

I can prepare a PR that is going to remove `gox` and use some approach like this http://dave.cheney.net/2015/08/22/cross-compilation-with-go-1-5 to build.

Or I can fork https://github.com/mitchellh/gox and prepare a PR (this approach takes long, but would be better for open source community) with the ability to use `GOARM`.

mattfarina:
@franciscocpg Before you fork gox see if there is interest there. Dealing with ARM properly has been overlooked numerous times.

What makes gox interesting is the parallel builds. Doing them in series may take too long.

franciscocpg:
Yes,  I thought about the parallelism too.

But even if gox doesn't have interest in the PR (I started the thread [here](https://github.com/mitchellh/gox/issues/67)) we can work with our fork, right?

Because we need that for sure.

mattfarina:
We can work with the fork.

franciscocpg:
So @mattfarina 
I created the PR https://github.com/Masterminds/glide.sh/pull/3 with the new version of `install.sh` and removed it from this PR.
ASAP I am going to work on gox.

mattfarina:
Now that the curl script is merged (and live) I'm going to close this.

We can handle the GOARM separately.

Issue #469, opened 2016-06-11
Bitbucket and subpackages: Cannot detect VCS


vburenin:
It happened so, that we are using bitbucket for our internal Git repository. It seems like glide has some problems with it.

Our internal repository is configured like this. All git ssh keys are available on the system.

Glide version:
glide version 0.11.0-dev (c25976f8059b0749f821d4e9c3faf7a205114097)

config
```yaml
- package: go-superlib
  repo: ssh://git@stash.local.net:7999/superlib/go-superlib.git
  vcs: git
```
Glide version: glide version 0.11.0-dev

During the run of 'glide up' command I am getting the following set of warnings and errors:
```text
[INFO]	Downloading dependencies. Please wait...
....
[INFO]	Fetching updates for go-superlib.
....
[INFO]	Resolving imports
[INFO]	Downloading dependencies. Please wait...
[INFO]	Fetching updates for go-superlib/lib1.
[INFO]	Fetching updates for go-superlib/lib2.
[WARN]	go-superlib/lib1 appears to be a vendored package. Unable to update. Consider the '--update-vendored' flag.
[INFO]	Fetching updates for go-superlib/lib3.
[INFO]	Fetching updates for go-superlib/lib4.
[INFO]	Fetching updates for go-superlib/lib5/lib6.
[WARN]	go-superlib/lib2 appears to be a vendored package. Unable to update. Consider the '--update-vendored' flag.
[WARN]	go-superlib/lib3 appears to be a vendored package. Unable to update. Consider the '--update-vendored' flag.
[WARN]	go-superlib/lib4 appears to be a vendored package. Unable to update. Consider the '--update-vendored' flag.
[WARN]	go-superlib/lib5/lib6 appears to be a vendored package. Unable to update. Consider the '--update-vendored' flag.
[ERROR]	Failed to set version on go-superlib/lib1 to : Cannot detect VCS
[ERROR]	Failed to set version on go-superlib/lib2 to : Cannot detect VCS
[ERROR]	Failed to set version on go-superlib/lib3 to : Cannot detect VCS
[ERROR]	Failed to set version on go-superlib/lib4 to : Cannot detect VCS
[ERROR]	Failed to set version on go-superlib/lib5/lib6 to : Cannot detect VCS
[INFO]	Project relies on 38 dependencies.
An Error has occurred
```
All these lib1,lib2... are just the internal packages which are the part of "superlib".

[...]

PR #470, opened 2016-06-11
Add a parameter to ignore subpackages check.

vburenin:
This is just a proposal that may help with fixing https://github.com/Masterminds/glide/issues/469

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -132,6 +132,15 @@ func (c *Config) MarshalYAML() (interface{}, error) {
 	return newConfig, nil
 }
 
+func (c *Config) IgnoreSubpackages(name string) bool {
+	for _, d := range c.Imports {
+		if d.NoSubpackages && strings.HasPrefix(name, d.Name+"/") {
+			return true
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


itcreator:
+1

franciscocpg:
I am getting the same error using an internal gitlab

vburenin:
So I guess, we could merge this fix when I add tests, however, I do not like how I named the new parameter.

"nosubpackages" kinda conflicts with existing "subpackages". Any better proposition?

sergeyklay:
+1

vovanec:
+1

vovanec:
ignore_subpackages?

yulyt:
+1

mattfarina:
First, thank you for putting some time into this. I appreciate the time put in and I appreciate that folks want this fixed.

A change to support package names that don't conform with `go get` should happen in name normalization. I would start inside [`util.NormalizeName`](https://github.com/Masterminds/glide/blob/f405e3789b5925d7282d4c2afab59a3827a688af/util/util.go#L306). If the regex lookup and go-get search don't find anything then handle this case.

PR #473, opened 2016-06-14
Fix indentation of example glide.yaml

rhcarvalho:
Make the indentation level for both `owners` the same.


mattfarina:
@rhcarvalho thanks for the fix!

PR #477, opened 2016-06-17
action: fix minor typo

gyuho:


mattfarina:
@gyuho Thanks for the fix!

gyuho:
np! thanks for the glide!

PR #480, opened 2016-06-20
Fix typo in config wizard

7imon7ays:
Edit 'major.minor.path' to be 'major.minor.patch'.


mattfarina:
@7imon7ays thanks for fixing that.

Issue #483, opened 2016-06-21
"<= 2.x is equivalent to <= 3"


derelk:
According to https://github.com/Masterminds/glide/blob/master/docs/versions.md#wildcards-in-comparisons:
> <= 2.x is equivalent to <= 3

Maybe I'm misunderstanding, but that doesn't sound right. That would imply `3.0` satisfies `<= 2.x`, no? Should maybe be `< 3`.

sdboyer:
Huh. That's an interesting case. Maybe I've forgotten some arcane rule that we're following with that, but just off the cuff, I'd tend to agree.

Could you open an issue against https://github.com/Masterminds/semver? That's where the fix would go.

derelk:
I don't have any reason to think this is an implementation error—I think it's a documentation error in this project. The documentation says:
> <= 2.x is equivalent to <= 3

But I think it should say:
> <= 2.x is equivalent to < 3

sdboyer:
Sorry, I wasn't clear - it IS an implementation error. Before writing my previous comment, I added a quick line to one of `Masterminds/semver`'s able tests to see if `3.0.0` is accepted by a `<=2.x` constraint. And lo, it is.

So the documentation is accurate with respect to how the semver lib works; we can't update glide's documentation until the issue is addressed over there.

derelk:
Got it, thanks.

mattfarina:
Is this a docs bug or implementation bug? <= 2.x should be equivalent to < 3.

sdboyer:
> it IS an implementation error.

sdboyer:
Actually, I'm not sure that it's a bug with semver's master branch - i did my test against 2.x

derelk:
This works as expected for me against semver 1.0.0, 1.0.1, and 1.1.0. I'm running:
```go
	c, _ := semver.NewConstraint("<= 2.x");
	v, _ := semver.NewVersion("3.0")
	fmt.Println(c.Check(v))
```
and it's consistently false. Thus I still believe this is simply a documentation error in this project.

derelk:
This is broken in the 2.x branch, though. The above code with `c.Matches(v)` returns `nil`.

sdboyer:
Ah cool, OK - so then it's probably fine to update it in the glide docs now, as the problem is just with `2.x`. And that's not primetime yet.

*glances sidelong at his TODO list*

derelk:
I'll file 2.x the issue in semver and a PR for this doc fix.

[...]

PR #488, opened 2016-06-23
fix for #483

derelk:
Per #483, "<= 2.x is equivalent to <= 3" should actually be "< 3" since
3.0 does not satisfy `<= 2.x`.


mattfarina:
@derelk thanks for the fix!

PR #489, opened 2016-06-23
Make shared funcs for lockfile usage

heewa:
Lightly refactor some lockfile logic into shared functions, so callers inside & outside glide do things the same way.

Personally I need this to create a glide plugin (as requested in PR #266), for which I need to read the lockfile, and turn those `Lock` structs into `Dependency` structs. These helper funcs will ensure that if anything changes in glide, my (and others') plugin will keep up.

(Also moving the `LoadLockfile` func has been a `TODO` in the code for a while)

### DIFF
--- a/action/install.go
+++ b/action/install.go
@@ -1,7 +1,6 @@
 package action
 
 import (
-	"io/ioutil"
 	"path/filepath"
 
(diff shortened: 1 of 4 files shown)


mattfarina:
@heewa thanks for the contribution!

PR #494, opened 2016-06-27
Glide install skips fetch when it is up to date

technosophos:
For 'glide install', this lazily computes whether the existing vendor
checkouts are already at the right commit. If they are, those deps are
skipped during the concurrent update.

This is safe to do in install because the dependency graph is already
computed, so there should never be a case where the glide.lock file does
not contain a negotiable dependency graph.

### DIFF
--- a/repo/installer.go
+++ b/repo/installer.go
@@ -111,8 +111,8 @@ func (i *Installer) Install(lock *cfg.Lockfile, conf *cfg.Config) (*cfg.Config, 
 
 	msg.Info("Downloading dependencies. Please wait...")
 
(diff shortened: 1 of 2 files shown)


repo/installer.go
```diff
@@ -302,6 +302,43 @@ func (i *Installer) List(conf *cfg.Config) []*cfg.Dependency {
 	return conf.Imports
 }
 
+// LazyConcurrentUpdate updates only deps that are not already checkout out at the right version.
```
sdboyer: ahh, saw this too late - s/checkout out/checked out/

PR #512, opened 2016-07-18
Fixed #505: Ignored packages can still be resolved

### LINKED ISSUE #505: Ignored dependency in testImports
I have a project with a `glide.yaml` file with the following (among other things):
```
ignore:
  - common
  - common/http
```

This project has a test file that imports `common/http`. After running `glide update` on v0.11.0 I see the following in my `glide.lock`
```
testImports:
- name: common/http
  version: ""
```

Basically an explicitly ignored module is showing up in the in `glide.lock`. I'm aware that doing relative imports in this way isn't the advised pattern. However - this was pretty unexpected behavior and I would guess is a bug? The reason I noticed it was that it broke the `glide install` I ran later on.

Hope this is helpful, let me know if any more clarification would be useful.

mattfarina:


charneykaye:
:shipit:

PR #522, opened 2016-07-26
Sort subpackages for glide.yaml and glide.lock

### LINKED ISSUE #521: Glide should sort subpackages alphabetically to avoid spurious diffs
I've noticed that the re-arrangement of `subpackages:` sections introduces changes to the lock file that are more just clutter than anything else. On a recent update of Kubernetes, I spent quite a bit of time just tracking movements inside of the lock file.

It seems to me that we could just sort subpackages lexicographically, and that would lead to predictable ordering. The easiest way to do this would be in a marshaling hook in the `cfg/lock.go` and `cfg/conf.go` files.

technosophos:
This alters the YAML marshaling to sort import and devimport subpackage
sections.

Closes #521

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/Masterminds/glide/util"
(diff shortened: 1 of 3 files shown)


technosophos:
@mattfarina I can't think of any cases where sorting like this would cause a problem. Can you?

franciscocpg:
Nice and KISS for sure!

mattfarina:
LGTM

PR #533, opened 2016-07-30
Log VCS error output as debug messages

atombender:
Enables logging of VCS error output when `--debug` is enabled.

I don't know if this is the ideal design — to be sure, the VCS error isn't ideal, and there's no central error reporting mechanism other than `msg`, which is why I extract the error there.

If I were to improve this, it'd be to encapsulate "some program failed to run and this is its failure info" as an error interface.

But I was fed up with getting unhelpful error messages ("Unable to get repository") from Glide, and this fixes that.

### DIFF
--- a/msg/msg.go
+++ b/msg/msg.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"strings"
 	"sync"
+
+	"github.com/Masterminds/vcs"
 )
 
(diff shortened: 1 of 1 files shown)


mattfarina:
Thanks for your contribution.

I understand your frustration. Thanks for helping to make it better.

PR #535, opened 2016-08-02
Resolve vendor directory symlinks

Fugiman:
If `vendor` is a symlink to another directory (for example, `__vendor`), `filepath.Walk` will not descend into it, preventing `--strip-vcs` and `--strip-vendor` from working. Appending `os.PathSeperator` is enough to resolve the symlink.

I'm not sure if this is the ideal solution, but I copied the pattern from [`glide-vc`](https://github.com/sgotti/glide-vc/blob/master/gvc.go#L117) which works as expected on symlink'd vendor directories.

### DIFF
--- a/path/path.go
+++ b/path/path.go
@@ -80,6 +80,25 @@ func Vendor() (string, error) {
 
 	gopath := filepath.Join(yamldir, VendorDir)
 
+	// Resolve symlinks
+	info, err := os.Lstat(gopath)
+	if err != nil {
+		return gopath, nil
+	}
+	for i := 0; IsLink(info) && i < 255; i++ {
+		p, err := os.Readlink(gopath)
+		if err != nil {
+			return gopath, nil
+		}
+
+		gopath = filepath.Join(filepath.Dir(gopath), p)
+
+		info, err = os.Lstat(gopath)
+		if err != nil {
+			return gopath, nil
+		}
+	}
+
 	return gopath, nil
 }
 
--- a/path/path_test.go
+++ b/path/path_test.go
@@ -34,6 +34,7 @@ func TestVendor(t *testing.T) {
 		t.Fatal(err)
 	}
 	wd, _ := os.Getwd()
+
 	os.Chdir(filepath.Join(td, "a/b/c"))
 	res, err := Vendor()
 	if err != nil {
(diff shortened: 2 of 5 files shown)


mattfarina:
Thanks for the PR. Before merging I'd like to understand what's going on here and why. Also, I want to make sure it works across nix/windows.

Fugiman:
Sure! I made https://github.com/Fugiman/vendortest to try and make this easier to reproduce. I ran `glide up -u -v -s` in that repo to generate the current state, and then ran `rm -rf vendor/**/.git` to ensure submodules weren't committed.

This repo demonstrates the current behavior of "`glide up -u -v -s` does not strip nested vendor directories if `vendor` is a symlink", and should be able to be used to ensure the patch works on nix/windows (I only tested on mac).

---

As for why appending `/` to `vendor` works: Glide uses `glide/path.StripVendor` to remove nested vendor directories. That uses `filepath.Walk` to recurse through the vendor directory. `filepath.Walk` users `os.Lstat` to determine if a file is a directory, which does not resolve symlinks. As far as I can tell, what `os.Lstat` does varies by platform. However, for mac appending `/` causes `os.Lstat` to return a `FileInfo` where `IsDir` is `true`.

I could make this behavior more explicit by manually resolving the symlink before passing it into `filepath.Walk` if you want?

mattfarina:
@Fugiman could you detect a symlink and resolve the location? The tailing `/` is something I don't really trust cross platform which we need to work. There is an [`IsLink`](https://github.com/Masterminds/glide/blob/6c2d90ce488abb132fa58ea234a941287730aefd/path/path.go#L165) function to detect symlinks. In the resolver there is also a function to [return the base directory for a symlink](https://github.com/Masterminds/glide/blob/31debc2ec4f36742ebb4b01f80e2a768586ace7c/dependency/resolver.go#L1052). These may be helpful.

Fugiman:
@mattfarina As requested, I now explicitly detect and resolve symlinks. I made the change to `Vendor()`, and then modified `StripVcs()` and `StripVendor()` to use that function.

I added another test to `Vendor()` to ensure the symlink behavior remains functional, and adjusted the scaffolding of the `StripVcs()` test to work with the format expected by `Vendor()`.

Please let me know if you'd prefer to have the change in the Strip functions rather than Vendor and I'll move it. I also set it up to resolve nested symlinks, but I don't know if that is overkill or not.

mattfarina:
@Fugiman thanks for the contribution!

PR #539, opened 2016-08-05
Add RPM spec file and version management to Makefile.

jollyroger:
This commit allows user to build an RPM package with a single command just
using source tarball. To achieve this i've added an sdist target to the
Makefile. This target will generate source tarball and put VERSION file inside
to allow make utility to detect version of the source tarball even if no git is
present.

If used a tarball generated by sdist target it is possible to build RPMs with a
single command:

    rpmbuild --define "_glide_version 0.11.1-11-gb5ebdcb" -ta \
        dist/glide-0.11.1-11-gb5ebdcb-src.tar.gz


mattfarina:
@jollyroger thanks for investing in RPM. Why do you think the RPM build should be part of the main glide repo? I ask because setup to [create debian package is in a separate repo](https://github.com/Masterminds/glide-ppa). If you can provide more insight it would be helpful.

jollyroger:
@mattfarina , I know that Debian does not prefer to use upstream-provided `debian/` directory and will repack the original tarball before putting it into the Debian archive. As for the RPM-based distros there is no such policy afaik. But since it's just a single file it's nice to have it somewhere nearby (haven't seen any packages of glide around except the Debian ones). When I needed an RPM and found none I just did packaging myself, additional changes resulted in the patch I provided earlier. 

Apart from the `.spec`-file itself, changes in the Makefile:

 * allow to package version-aware source distribution tarball (which is great for any packaging solution)
 * allow to install additional files common for packages (license, readme)
 * support PREFIX variable commonly used during packaging to put binary files into `/usr` instead of `/usr/local`
 * support to redefine `VERSION`, `PREFIX` variables for the purposes of packaging

So if your concern is for spec-file only and the rest seems ok to you, I can split my patch into two.

jollyroger:
Friendly bump.

mattfarina:
@jollyroger I'll look at this shortly. I need to get back up to speed on RPMs before making a decision. This needs to be forward compatible since as we continue working (it may already be and I'm just not aware).

jollyroger:
@mattfarina thanks. Will wait for your answer. Didn't understood your concern though. Could you please put in more details? I'll gladly answer whether this solution already fits your requirements or not.

leprechau:
For what it's worth I maintain an AlpineLinux package for glide that's been accepted upstream.

jollyroger:
Friendly bump.

Issue #440, opened 2016-05-24
glide tree error "panic: runtime error: invalid memory address or nil pointer dereference"


gubei:
When using glide tree to print dependency tree , I found a error.

 |       |-- github.com/go-sql-driver/mysql   (/home/gubei/gocode/src/github.com/xiaohuanggua/virgin/vendor/github.com/go-sql-driver/mysql)
|       |       |-- appengine/cloudsql   ()
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x20 pc=0x4e2233]

goroutine 1 [running]:
panic(0x945080, 0xc82000e170)
        /usr/local/Cellar/go/1.6/libexec/src/runtime/panic.go:464 +0x3e6
github.com/Masterminds/glide/dependency.srcDir(0x0, 0x0, 0xc81ffea1e0)
        /Users/mfarina/Code/go/src/github.com/Masterminds/glide/dependency/resolver.go:927 +0x33
github.com/Masterminds/glide/dependency.IsSrcDir(0x0, 0x0, 0xc81ffea1e0)
        /Users/mfarina/Code/go/src/github.com/Masterminds/glide/dependency/resolver.go:923 +0x2b
github.com/Masterminds/glide/tree.walkDeps.func1(0x0, 0x0, 0x0, 0x0, 0x7fa5ea9a1608, 0xc8202bc3f0, 0x0, 0x0)
        /Users/mfarina/Code/go/src/github.com/Masterminds/glide/tree/tree.go:48 +0x92
path/filepath.Walk(0x0, 0x0, 0xc8201ccac8, 0x0, 0x0)
        /usr/local/Cellar/go/1.6/libexec/src/path/filepath/path.go:394 +0x

my os system is fedora linux .

thank you for helping.

abhin4v:
I get the same error on OSX 10.10.5:

    $ ~/.gocode/src/github.com/Masterminds/glide/glide tree
    [... some output ...]
    goroutine 1 [running]:
    panic(0x5723a0, 0xc8200120e0)
           	/usr/local/Cellar/go/1.6/libexec/src/runtime/panic.go:464 +0x3e6
    github.com/Masterminds/glide/dependency.srcDir(0x0, 0x0, 0xc81ffe5553)
           	/Users/abhinav/.gocode/src/github.com/Masterminds/glide/dependency/resolver.go:1035 +0x33
[...]

PR #544, opened 2016-08-10
Fixes panic while running tree command

abhin4v:
Checks for error while walking directories and propagates them up to prevent panic. Fixes issue #440.


mattfarina:
@abhin4v thanks for the contribution

Issue #553, opened 2016-08-18
export step returning error with docker


adamwilbert:
[ERROR]	Unable to export dependencies to vendor directory: rename /tmp/glide-vendor378089091/vendor /go/.../vendor: invalid cross-device link

This is causing an error in my build step, formerly fetching/setting worked fine, but now exporting step is creating an issue.

mattfarina:
Can you test the PR at #556.

mattfarina:
Merged PR. This is fixed.

PR #556, opened 2016-08-18
#553: Fixed issue with moving vendor directory cross devices with copy

mattfarina:


eriklincoln:
Works for me with installing it with:

```yaml
      - go get github.com/Masterminds/glide
      - cd $GOPATH/src/github.com/Masterminds/glide
      - git checkout fix/553
      - git pull origin fix/553
      - go install github.com/Masterminds/glide
      - cd $DRONE_DIR
      - install deps, build, etc.  
```

PR #574, opened 2016-08-22
Use msg.Debug instead of fmt.Println in the case of hash mismatch.

albrow:
I was seeing the entire __glide.yaml__ file printed out on our CI server and it was distracting. This output is really verbose and should probably only be displayed if the `--debug` flag was provided. Are you okay with changing this to `msg.Debug`?

### DIFF
--- a/action/install.go
+++ b/action/install.go
@@ -1,7 +1,6 @@
 package action
 
 import (
-	"fmt"
 	"path/filepath"
 
 	"github.com/Masterminds/glide/cache"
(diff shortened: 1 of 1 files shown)


mattfarina:
@albrow thanks for pointing this out. That debug code should all be removed. It was left in by accident.

Because I'm releasing today (if all things go well) I'll make this fix. Thanks so much for catching this.

albrow:
@mattfarina I was wondering if that was the case. Thanks for the quick response!

Issue #568, opened 2016-08-22
Remove verbose flag as it does nothing


mattfarina:

mattfarina:
This was merged in #582

PR #582, opened 2016-08-25
Remove verbose flag

kelcecil:
The verbose logging in the messenger isn't used anywhere, so this removes the command line
flag as well as cleaning up additional unused functions and fields. This is intended to address issue #568.


mattfarina:
Thanks. Once I switch focus to 0.13 I'll merge this.

PR #587, opened 2016-08-29
Fixed #580: cross-device error handling failed on Windows

### LINKED ISSUE #580: The system cannot move the file to a different disk drive
On Windows, glide can only be used on the same disk drive that the default cache directory is located:

```
# glide install

[INFO]  Downloading dependencies. Please wait...
[INFO]  --> Found desired version locally ...
[INFO]  Setting references.
[INFO]  --> Setting version for ...
[INFO]  Exporting resolved dependencies...
[INFO]  --> Exporting ...
[ERROR] Unable to export dependencies to vendor directory: rename C:\Users\MyUser\AppData\Local\Temp\glide-vendor439134107\vendor E:\workspace\src\...\vendor: The system cannot move the file to a different disk drive.
```

Workaround: create a folder outside your GoPath, but on the same disk, and pass that folder to glide via `glide --tmp [folder]`.

Environment: Windows 7 32-bit (MinGW bash shell), Go 1.7, Glide v0.12.0.

mattfarina:


repo/installer.go
```diff
@@ -360,8 +361,11 @@ func (i *Installer) Export(conf *cfg.Config) error {
 	err = os.Rename(vp, i.VendorPath())
 
 	// When there are different physical devices we cannot rename cross device.
-	// Fall back to manual copy.
-	if err != nil && strings.Contains(err.Error(), "cross-device link") {
+	// Note, windows does not return the cross-device link message but instead
+	// bubbles up a system message (I believe is i18n). So, we try to detect the
+	// cross device link error or a windows error. If copy fails that will bubble
+	// up an additional error.
+	if err != nil && (strings.Contains(err.Error(), "cross-device link") || runtime.GOOS == "windows") {
```
sdboyer: if it's going to sniff the err for a particular string on *nix, shouldn't it sniff for windows, too?

mattfarina: @sdboyer The error from Windows comes from the system and is multi-lingual. Any suggestions on how to detect it in all possible languages?

jrick: I believe the `*LinkError` returned by `Rename` should have the `Err` field set to a `syscall.Errno`, which is a uintptr and that can compare against a Win32 error code.

sdboyer: thaaat sounds much better. would be better to do the same kind of check for other arches, too

mattfarina: This does sound much better. I'll work on that today.

mattfarina:
The latest commit does a few changes. First, instead of parsing the string (which can vary) it checks for the syscall error by name. Then, on Windows it falls back to check the system error if that was returned.

PR #591, opened 2016-08-31
Fixed #590: When exit signal recieved remove global lock

### LINKED ISSUE #590: lock.json not cleaned up on signals
When I ctrl-c glide while running the `~/glide/lock.json` file is not cleaned up. It probably should be. 

I know glide resolve the file already existing itself, but it would be good to handle this normal case.

mattfarina:


cache/global_lock.go
```diff
@@ -81,6 +82,23 @@ func startLock() error {
 		}
 	}()
 
+	// Capture ctrl-c or other interruptions then clean up the global lock.
+	ch := make(chan os.Signal)
+	signal.Notify(ch, os.Interrupt, os.Kill)
+	go func(cc <-chan os.Signal) {
+		s := <-cc
+		SystemUnlock()
```
chancez: Seems racy to me. Nothing else gets blocked on the signal, so after you unlock it, another go routine could potentially acquire (create) a lock after this call, and before the exit.

mattfarina: Good catch. There is a small window. Fixing...

PR #617, opened 2016-09-20
Glide Cache Path

akutz:
This patch introduces a new command, `cache`. The `cache` command has two sub-commands: `clear` and `path`. The `cache clear` command is simply the former `cache-clear` command relocated under the new `cache` command. The `path` command may be used to return the location of a dependency's sources in the Glide cache. For example:

```bash
$ cd $GOPATH/src/github.com/emccode/rexray && \
  glide cache path github.com/emccode/libstorage
```

The above command would return the following location:

`$HOME/.glide/cache/src/https-github.com-emccode-libstorage`


mattfarina:
@akutz this looks fairly straight forward. Sorry I just got around to it today.

Overall it looks good.

As to not merge a change right before the weekend, I'll tackle this on Monday.

action/cache.go
```diff
@@ -21,3 +23,17 @@ func CacheClear() {
 
 	msg.Info("Glide cache has been cleared.")
 }
+
+// CachePath returns the path to a cached dependency's sources.
+func CachePath(importPath string) {
```
sdboyer: This isn't actually returning anything?

akutz: I amended the commit to reflect the proper comment. It was written before I realized your functions prefer to print/die instead of return.

akutz:
Hi @mattfarina / @sdboyer,

Any additional thoughts on this? Thanks!

akutz:
Hi @mattfarina / @sdboyer,

Any additional thoughts on this? Thanks!

glide.go
```diff
@@ -719,12 +719,33 @@ Example:
...
+				{
+					Name:  "path",
+					Usage: "Get a the path to a cached dependency's sources",
+					Description: `Returns the path to a cached dependency's
+		   sources. This is useful for scripts that need to run commands like
+		   "git-describe" on one of a project's vendored dependencies.
+
+		   For example,
+
+		       glide cache path github.com/emccode/libstorage`,
```
mattfarina: Can you add a note that these are the dependency locations in the glide.yaml or glide.lock file? So that one knows what arguments should be passed in.

action/cache.go
```diff
@@ -21,3 +23,17 @@ func CacheClear() {
 
 	msg.Info("Glide cache has been cleared.")
 }
+
+// CachePath prints the path to a cached dependency's sources.
+func CachePath(importPath string) {
+	for _, d := range EnsureConfig().Imports {
```
mattfarina: `EnsureConfig` doesn't load the lock file. That means it won't find transitive dependencies. It might be a good idea to load a lock, if present, and fall back to config if the lock is not there.

PR #627, opened 2016-10-02
Catch back up to master

sdboyer:
This just resolves a number of standing conflicts between `gps-integration` and `master`. There may be some issues introduced by this; I tried to check through as thoroughly as I could, but I didn't realize it'd been two months since I'd caught my branch up to master. Lotsa changes.

Probably most notable here is that there's no support for v0.12.x's new mirroring system.


mattfarina:
@sdboyer we'll need to fix mirroring before we merge back into master. I'd really like the 0.13.x releases to only have gps as the change. so as to not complicate things.

sdboyer:
@mattfarina right, this is against the `gps-integration` branch, not master. it merges IN master in order to resolve conflicts - something that, IMO, we should be doing as we go in order to minimize conflicts and integration work once we finish other things in the checklist.

no?

mattfarina:
@sdboyer I ran `go run glide.go up` twice on this branch. On the second run I got errors. You can see the second run in the [gist here](https://gist.github.com/mattfarina/4465b759406a196aac1b9375e7963783). Can you talk to that?

sdboyer:
just updating here what we discussed in gitter - that issue may be related to a known problem in gps with concurrent fetches. when @mattfarina re-ran the command, the error didn't re-occur.

I've opened sdboyer/gps#106 to deal with that underlying bug.

Issue #629, opened 2016-10-04
0.12.3 not on master


breerly:
I see 0.12.2 on master, but not 0.12.3, even though it's released:

https://github.com/Masterminds/glide/releases/tag/v0.12.3

Possible to get that on master?

breerly:
For the 0.12.3 tag, version.go still shows `var version 0.12.0-dev`...

https://github.com/Masterminds/glide/blob/84607742b10f492430762d038e954236bbaf23f7/glide.go#L37

:(

breerly:
@mattfarina what is actually on master?

franciscocpg:
@breerly, maybe this is useful?
https://github.com/Masterminds/glide/issues/616

mattfarina:
@breerly is the tag not being on master a problem?

`glide.go` on master should be bumped. Good catch. I'll fix that.

There is a release branch for 0.12.x changes to keep with semver while we work on master with breaking changes.

breerly:
@mattfarina I was expecting Glide to do something like:

1. Latest release is always on master
2. Releases are made *off of* master
3. Changes for the next release aggregate in another branch
4. That branch gets merged into master when ready for a new release

#631 should keep us on track going forward.

PR #631, opened 2016-10-05
Verify version during build

breerly:
Let's verify that the `glide.Version` matches the changelog - this will force us to update this value before releasing and prevent #629 from happening again.

To use this correctly, you must:

1. The latest version listen in the CHANGELOG must match `glide.Version`.
2. `-dev` can optionally be added to `glide.Version`.
3. The latest unreleased version must be listed in the CHANGELOG.

We've used this strategy successfully in:

- https://github.com/yarpc/yarpc-go
- https://github.com/thriftrw/thriftrw-go

cc @mattfarina

### DIFF
--- a/glide.go
+++ b/glide.go
@@ -34,6 +34,8 @@ import (
 	"os"
 )
 
+const version = "0.13.0-dev"
+
 const usage = `Vendor Package Management for your Go projects.
 
    Each project should have a 'glide.yaml' file in the project directory. Files
@@ -58,7 +60,7 @@ func main() {
 	app := cli.NewApp()
 	app.Name = "glide"
 	app.Usage = usage
-	app.Version = Version
+	app.Version = version
 	app.Flags = []cli.Flag{
 		cli.StringFlag{
 			Name:  "yaml, y",
--- a/glide.go
+++ b/glide.go
@@ -34,8 +34,6 @@ import (
 	"os"
 )
 
-var version = "0.13.0-dev"
-
 const usage = `Vendor Package Management for your Go projects.
 
    Each project should have a 'glide.yaml' file in the project directory. Files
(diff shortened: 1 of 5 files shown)


breerly:
This first build should fail, verifying the feature. Once it does, I'll update the value of `glide.Version` to make it pass.

version.go
```diff
@@ -0,0 +1,4 @@
+package main
+
+// Version is the current version of Glide
+const Version = "0.13.0-dev"
```
prashantv: can unexport this

breerly: Might be useful for exterior libs to be able to get at this, in the case that they use glide as a lib.

prashantv: I'd prefer to keep the visibility the same as it was previously as part of this change.

However since this is a main package, other's can't import it anyway.

sdboyer:
i haven't thought through all the implications of this, but in general it seems like a great idea

breerly:
@sdboyer the implications is that the build will fail if `go.version` do not match the latest version in the changelog. In practice, this means catching issues like #629 from happening.

sdboyer:
@breerly yes, that i understand - that's the intended effect. when i use "implication," i'm referring to potential second-order, unintended effects.

this isn't a knock against the PR; it's just a caveat about the depth of my review.

glide.go
```diff
@@ -34,7 +34,7 @@ import (
 	"os"
 )
 
-var version = "0.13.0-dev"
+const version = "0.13.0-dev"
```
mattfarina: By changing from a `var` to `const` the `ldflags` for `-X main.version=${VERSION}` no longer works in the `make build`. This is why it was a `var` in the first place.

I think this needs to be switched back.

mattfarina:
Once `make build` can pass the version into the code again this looks ready.

I like having this check.

breerly:
Got it, I'll switch back to `var`

glide.go
breerly: Sure

mattfarina:
@breerly thanks for the contribution.

PR #642, opened 2016-10-19
fix replacing existing vendor dependencies

bassamtabbara:
When replacing existing vendor dependencies do not remove
the vendor directory. Instead just remove the contents
of the directory. This supports scenarios where the
vendor directory is bind mounted, and avoids "device busy"
error.


bassamtabbara:
This requires a way to copy whole directories too. Closing for now.

PR #686, opened 2016-11-22
Make ending dots in prints more consistent

stevenroose:
It seems that mostly sentences that indicate an action ("Xing something") do not end in a dot. But some did. So it's either all of them or none: I changed it to none, dots at the end of these messages look strange.


mattfarina:
@stevenroose thanks for the contribution and fixing the inconsistency.

PR #699, opened 2016-12-02
Try to preserve vendor/.git, if it exists

### LINKED ISSUE #697: `glide {get,install,update}` breaks `vendor` submodule
In our vendor workflow, we opted to make `vendor` a submodule, so we'd get the benefits of _checking in_ deps (e.g. maintaining distribution, reproducibility, etc) without adding the actual content of dep changes to our main repo history (our `vendor` is ~200mb right now).

However, glide _replaces_ the `vendor` directory with a newly built one after every operation, which has the side effect of deleting the `vendor/.git` submodule pointer.

Our current [workaround](https://github.com/cockroachdb/cockroach/blob/master/scripts/glide.sh) is to just move `.git` aside and replace it after.

sdboyer:
Fixes #697.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release 0.13.0 (unreleased)
 
-- Nothing yet
+- #697: Preserve vendor/.git, if it exists.
 
 # Release 0.12.3 (2016-10-03)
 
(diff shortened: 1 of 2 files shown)


sdboyer:
This works in my testing...but I had to abstract out the cross-link error handling a bit, and I don't have a suitable test environment at hand in which to test that that's working as expected.

sdboyer:
set up an env, we're good.

Issue #592, opened 2016-08-31
[Q] Clarifications on updating dependencies


a13xb:
Your usage example states:

```
$ glide up                                # Update to newest versions of the package
```

However, it doesn't seem to do that at all. If I start with a clean repository with `glide.yaml`, where none of the packages are constrained, and a matching `glide.lock`, none of the dependencies are updated, even though upstream has new commits. Repository remains clean.

When (pre-0.12) Glide was still not stripping VCS files, I used to be able to go into `venvor` tree, checkout a specific version of a single dependency and run `glide update`, it would update the lock file to pin the dependency at the version I checked out. So I've been left with the impression that all `glide update` does is transfer the state of your `vendor` tree into `glide.lock`, that's it.

Obviously now `vendor` tree doesn't retain VCS information, I'm not sure what exactly `glide update` is supposed to do, and how to actually properly update dependencies to the latest version of the upstream.

mattfarina:
I'm a little confused about something and I think I need some more information to understand what's going on.

The `glide.yaml` file holds the intent while the `glide.lock` file holds the specific revisions being pinned to. So, a `version` for a dependency in the `glide.yaml` file that's empty, tracks a branch, or has a version range causes Glide to go to the dependency source, pull down any updates, and store them locally. It does that automatically.

There was never the intent for someone to manually checkout a version inside the repo within the `vendor/` folder. Glide tells the VCS (such as Git) to go out to the source and fetch updates.

This happens via the [`github.com/Masterminds/vcs`](https://github.com/Masterminds/vcs) package. For example, with Git it will drop down to `git` and use commands like `git fetch` and `git pull` expecting them to go out.

Is that not happening for you?

a13xb:
> Is that not happening for you?

When I run `glide update`, yes, it pulls the latest from the upstream VCS into cache, but the versions in the `vendor/` workspace are not updated.

Let me clarify again: my `glide.yaml` is a plain list of dependencies, none of them specify a branch, a tag, a commit, or range of versions, none whatsoever. My expectation was that this will result in `glide update` always updating the `glide.lock` to the HEAD of master.

This doesn't seem to be happening. After the update of local cache, my `glide.lock` does not change, even though there are new commits in some upstream repos.
```
[INFO] 	Replacing existing vendor dependencies
[INFO] 	Versions did not change. Skipping glide.lock update.
```

[...]

PR #703, opened 2016-12-04
Do not rely on local git state in checkouts

sdboyer:


jmazzitelli:
@sdboyer I'm +1 on merging this. In the documentation you should mention that you must specify a "version" or "ref" field if going back and forth between branches, tags, commit hashes. See my comment in the issue #592 but I think you know what I mean :)

flimzy:
Any chance of getting this merged?

PR #704, opened 2016-12-04
add default timeout for resolving root from go get

YueHonghui:
In function ```getRootFromGoGet```, glide fire http get to resolve root from go get. This will be hanged for long time with some import path without "https" service or be banned by GreetFireWall in china. Suggest to add timeout into http request.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/Masterminds/vcs"
 )
@@ -21,6 +22,9 @@ import (
 // This is not concurrently safe which is ok for the current application. If
 // other needs arise it may need to be re-written.
 var ResolveCurrent = false
+var goGetCli *http.Client = &http.Client{
+	Timeout: 5 * time.Second,
+}
 
 func init() {
 	// Precompile the regular expressions used to check VCS locations.
@@ -84,7 +88,7 @@ func getRootFromGoGet(pkg string) string {
 		u.RawQuery = u.RawQuery + "&go-get=1"
 	}
 	checkURL := u.String()
-	resp, err := http.Get(checkURL)
+	resp, err := goGetCli.Get(checkURL)
 	if err != nil {
 		addToRemotePackageCache(pkg, pkg)
 		return pkg


franciscocpg:
IMO timeout should be an environment variable with some default value (like the 5 seconds)

YueHonghui:
update the PR with global option go-get-timeout

glide.go
```diff
@@ -87,6 +88,11 @@ func main() {
 			Usage:  "The temp directory to use. Defaults to systems temp",
 			EnvVar: "GLIDE_TMP",
 		},
+		cli.UintFlag{
+			Name:  "go-get-timeout",
```
sdboyer: Please name the parameter in a way that describes what it does. Its provenance is not important in naming.

glide.go
```diff
@@ -87,6 +88,11 @@ func main() {
 			Usage:  "The temp directory to use. Defaults to systems temp",
 			EnvVar: "GLIDE_TMP",
 		},
+		cli.UintFlag{
```
sdboyer: Please name the parameter something that describes its purpose. `http-timeout` would be fine.

util/util.go
```diff
@@ -20,7 +21,10 @@ import (
 // the current OS/ARCH instead of all possible permutations.
 // This is not concurrently safe which is ok for the current application. If
 // other needs arise it may need to be re-written.
+const DefaultGoGetTimeout time.Duration = time.Second * 5
+
 var ResolveCurrent = false
+var GoGetTimeout = DefaultGoGetTimeout
```
sdboyer: Please name the variable something related to its purpose. The provenance of this timeout is not relevant when it comes to variable naming.

YueHonghui:
I think the solution is much more complicated than initially thought. ```http-timeout``` is not the suitable name. If we use the name ```http-timeout```, that seems to set timeout for all http request in glide. It is difficult to select such timeout value for different size of response body. Maybe we should provide a timeout option for all http request just include time connecting and sending request, or just provide different options for every use case such as ```go-get-timeout```. Any suggestions?

sdboyer:
Fair criticisms for a general name like `http-timeout`, but `go-get-timeout` is still not a good enough name. We're not invoking `go get` under the hood, which is the most reasonable interpretation of that name.

PR #710, opened 2016-12-10
Allow import of dependencies from govendor

radeksimko:
This will allow import of existing dependencies tracked via [govendor](https://github.com/kardianos/govendor).

### Transitive dependencies

One caveat I bumped into while testing this out with [`hashicorp/terraform`](https://github.com/hashicorp/terraform) is that govendor (as opposed to glide) allows transitive dependencies to be pinned to different versions to what the original 1st level dependency requires.

I couldn't find any real benefit of that feature - in fact I think it's better if this treated as a conflict - like it is in Glide.

To solve this in most graceful way I decided to generate initial `glide.yaml` file which may not be error-less for the first time, but will include *all* dependencies and 1st `glide update` will report any conflicts and let the user resolve these, e.g.

```
[ERROR]	Failed to parse /Users/radek/gopath/src/github.com/hashicorp/terraform/glide.yaml: Import github.com/fsouza/go-dockerclient repeated with different versions 'bf97c77db7c945cbcdbf09d56c6f87a66f54537b' and 'b21e1152923f46354476a3befa5b7aab0f441b4c'
```

### `Version` vs `VersionExact`

There are 2 main fields that may define the pinned version. `Version` is typically more loose (`v1` which will match latest `v1.*.*` tag or branch) and `VersionExact` is exact match (typically resolved `Version`). I decided to give priority to `VersionExact` as I came to the conclusion that user is more likely to benefit from dependency versions that are as close to previous versions as possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ install: build
 	install -m 755 ./glide ${DESTDIR}/usr/local/bin/glide
 
 test:
-	${GLIDE_GO_EXECUTABLE} test . ./gb ./path ./action ./tree ./util ./godep ./godep/strip ./gpm ./cfg ./dependency ./importer ./msg ./repo ./mirrors
+	${GLIDE_GO_EXECUTABLE} test . ./gb ./path ./action ./tree ./util ./godep ./godep/strip ./govendor ./gpm ./cfg ./dependency ./importer ./msg ./repo ./mirrors
 
 integration-test:
 	${GLIDE_GO_EXECUTABLE} build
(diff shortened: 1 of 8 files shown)


sdboyer:
oh oh, it's awesome that you did this! this has been sitting around some TODO list of mine for a while.

i'll have a look at the implementation as soon as i get a chance (i hope later this week). yes, the difference between glide and govendor wrt listing packages vs. projects has been a significant point of discussion for a while; i've also had a couple thoughts on the best way to implement.

radeksimko:
Assuming the de-facto standard is now go modules (at least for projects that I work on) I'm going to close this PR, but I am happy for anyone else to pick it up, if they still have a need for 3rd-party dependency manager.

PR #717, opened 2016-12-16
Cache GOROOT at init time

heyitsanthony:
When GOROOT is not set as an environment variable, GetBuildContext
fetches it from 'go env GOROOT', which takes about 50ms on my machine.
This overhead can almost double the time it takes to run an update since
GetBuildContext is called relatively frequently.

Instead, read GOROOT during initialization and cache it for
GetBuildContext.


mattfarina:
@heyitsanthony thanks for the contribution!

PR #739, opened 2017-01-12
Fix #736: Find home dir without using cgo

### LINKED ISSUE #736: Bug: Unreliable code for resolving glide home
The following code, which resolves the location of the glide home when the `GLIDE_HOME` environment variable is not set, is unreliable. It uses `os/user`, which in turn uses cgo, which doesn't work when cross-compiling, which I assume you must be doing.

https://github.com/Masterminds/glide/blob/master/path/path.go#L40-L62

The result is that glide home resolves as documented on a Mac, but on a Linux machine (which, of course, includes any Docker container), the logic above falls back to resolving the glide home as `filepath.Join(cwd, ".glide")`.

While that's not entirely inconvenient, it is, to say the least, surprising and undocumented OS-specific behavior.

This could be corrected by abandoning the use of `os/user` in favor of https://github.com/mitchellh/go-homedir, which does not rely on cgo.

@mattfarina @technosophos I'd be happy to submit a PR for this if you agree with my assessment of the problem and my proposed solution.

krancour:
Fixes #736 

cc @mattfarina

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,12 +1,14 @@
-hash: 67c5571c33bfcb663d32d2b40b9ce1f2a05a3fa2e9f442077277c2782195729c
-updated: 2016-08-11T14:22:17.773372627-04:00
+hash: e36c5a71bd15a6d2f7f795a68b0fbfbd41537ee29aec16e7105bd8fd3ef118ee
+updated: 2017-01-12T13:12:10.704271522-05:00
 imports:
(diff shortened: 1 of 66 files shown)


mattfarina:
@krancour thanks for the contribution. I'm going to work through Glide changes for a bit and release a new version prior to bringing in GPS.

PR #740, opened 2017-01-15
Support importing of dependencies from Trash

stffabi:
This PR brings support for importing of existing dependencies tracked via [Trash](https://github.com/rancher/trash).


hferentschik:
Any ETA when this will be merged?

PR #747, opened 2017-01-24
Fix glide home directory

agatan:
`path.Home()` does not return user's home directory even if it exists.
I think this is because incorrect `err` checking.


mattfarina:
Thanks for the fix.

PR #757, opened 2017-02-02
Add glide-pin plugin

darkliquid:
Adds a link to the glide-pin plugin for pinning all discovered dependencies in the `glide.lock` file explicitly in the `glide.yaml` file. Respects version ranges, but will replace them if the `--exact` flag is used.


mattfarina:
Thanks for the contribution.

PR #800, opened 2017-04-10
Build: use VERSION if provided or fallback to git tag --describe

BlackYoup:
Hi,

I'm packaging this software for a source based distribution [(Exherbo)](https://exherbo.org/). During the compilation, we are not in a git repository (I use Github tarball). Because of that, the `$(shell git tag --describe`) will emit an error even if I pass the `VERSION` to `make` like: `make build VERSION=0.12.3`

The error is harmless and the build still succeed but it can make the user think that the compilation will fail :-)


mattfarina:
Thanks for the contribution.

PR #812, opened 2017-04-22
GLIDE-811: Update glide to use vcs w Windows git submodule fix

m0j0hn:
DO NOT MERGE THIS PR!
This PR is intended to exercise / test a patched vcs in Glide -
the proper fix will be to merge the vcs patch, and then update Glide to use the newer version
of vcs - which *will* require a PR, just not this one, probably.


mattfarina:
@m0j0hn FYI, in the future you can use the `repo` property in the `glide.yaml` file to alias to a different repo instead of rewriting paths.

Closing because #816 is in place.

m0j0hn:
Awesome! Thanks, @mattfarina !!!
(and thanks for the tip about `repo` - I think I didn't understand how it could be used like this, very useful!)

PR #829, opened 2017-05-18
Preserve .git file correctly

### LINKED ISSUE #828: glide screws up `.git` file when using a submodule
I ran into an issue that was very frustrating to track down.

The `vendor` directory is a submodule, so it has a `.git` file (not a directory):
```
# cat vendor/.git 
gitdir: ../.git/modules/vendor
```

After running `glide up`, this is replaced with an empty directory with no execute permissions:
```
# ls -la vendor/.git
ls: cannot access 'vendor/.git/..': Permission denied
ls: cannot access 'vendor/.git/.': Permission denied
total 0
d????????? ? ? ? ?            ? .
d????????? ? ? ? ?            ? ..
```

This bogus directory causes `git status` to fork-bomb my machine (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=750687). The icing on the cake is that I have my shell run `git status` automatically to show some info in the prompt :)

I looked at the code a bit and I believe the culprit is this call: 
https://github.com/Masterminds/glide/blob/b51014108d43cc2c4aa907584f2a68e72dc3b708/repo/installer.go#L369

Indeed, on my system `/tmp` is a separate filesystem, so the rename wouldn't work. This code should be fixed so it copies the file (if it's just a file).

RaduBerinde:
When vendor is a submodule, .git is a file (not a directory). When /tmp is on a
different filesystem, we try to copy it as a directory which leaves an empty
.git directory instead of the file. This change fixes this by copying it as a
file in this case.

Fixes #828.

---
This change is [Reviewable](https://reviewable.io/reviews/masterminds/glide/829)

### DIFF
--- a/repo/installer.go
+++ b/repo/installer.go
@@ -359,16 +359,23 @@ func (i *Installer) Export(conf *cfg.Config) error {
 	// move it over to the newly-generated vendor dir - the user is probably
 	// submoduling, and it's easy enough not to break their setup.
 	ivg := filepath.Join(i.VendorPath(), ".git")
-	_, err = os.Stat(ivg)
+	gitInfo, err := os.Stat(ivg)
 	if err == nil {
-		msg.Info("Preserving existing vendor/.git directory")
+		msg.Info("Preserving existing vendor/.git")
 		vpg := filepath.Join(vp, ".git")
 		err = os.Rename(ivg, vpg)
 
 		if terr, ok := err.(*os.LinkError); ok {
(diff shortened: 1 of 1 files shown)


dt:
@sdboyer did you get a chance to look at this in between all the dep work?

sdboyer:
@dt hah, i didn't even see it, thanks for the ping.

I totally never knew that submodules did this. Huh. Since 1.7.8, it seems...I wonder how I missed it for so long.

Anyway, LGTM - thanks!

PR #846, opened 2017-06-11
reduce install dependencies to just Go

mcandre:
Remove the need for curl and similar coreutil tools when installing Glide


jbergstroem:
might want to mention that go bin path needs to be in path as well

PR #857, opened 2017-06-29
Fixed #778: Removing and moving large sets of files fails on Windows

### LINKED ISSUE #778: bash on windows has error
I use glide in bash on windows. when i exec `glide install`.i got the follow error message
```shell
[INFO]  --> Exporting golang.org/x/net
[INFO]  --> Exporting golang.org/x/crypto/acme/autocert
[INFO]  --> Exporting golang.org/x/crypto
[ERROR] remove /tmp/glide-vendor220918663/vendor/github.com/google/go-github/github: directory not empty
[ERROR] Unable to export dependencies to vendor directory: remove /mnt/c/Users/Administrator/go/src/github.com/freshcn/go-id-builder/vendor/github.com/google/go-github/github: directory not empty
```
there has the same error on  v0.12.3 and 0.13.0-dev

mattfarina:
For an example of the issue and status on Go itself see
https://github.com/golang/go/issues/20841

When working with large file sets such as those in a large dependency
tree (e.g. thousands of files) functions such as os.RemoveAll and
os.Rename can fail on Windows and Windows Subsystem for Linux. The
system commands work without issue.

This change detects Windows and the WSL environments then drops
down to system commands to make these changes. On mac, linux, and
other the os package functions are still used.

### DIFF
--- a/path/strip.go
+++ b/path/strip.go
@@ -30,7 +30,7 @@ func StripVendor() error {
 			if _, err := os.Stat(path); err == nil {
 				if info.IsDir() {
 					msg.Info("Removing: %s", path)
-					return os.RemoveAll(path)
+					return CustomRemoveAll(path)
 				}
 
 				msg.Debug("%s is not a directory. Skipping removal", path)
--- a/path/winbug.go
+++ b/path/winbug.go
@@ -1,0 +1,96 @@
+package path
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime"
+
+	"bytes"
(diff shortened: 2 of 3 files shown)


djensen47:
It works for me on Windows Subsytem for Linux.

I was never able to reproduce on Windows proper so somebody else will have to try that.

kumarharsh:
It works for me on Windows proper. Have given this only one shot - but it worked without any error, whereas the glide.exe taken from releases didn't. I'll give it more tries later tomorrow. 👍

Issue #652, opened 2016-10-24
Glide vendors all packages, nut just specified subpackages


karalabe:
When only a subpackage is used from a larger one, glide will pull in the entire repository, even though the subpackage is explicitly specified in the .yaml file. This results in tons of new dependencies and code being pulled in.

E.g. A priority queue library pulls in the entire QT because there's some example on a completely different repository path that happens to use WebEngine.

AaronO:
@karalabe Not a fix per say, but I built a small tool called `go-vendor-clean`, it's not yet open source, but I use it in conjunction with `glide` to get glide's versioning but lean/small `vendor/` folders.

`go-vendor-clean` supports `glide`, `godep` and still works if you have neither.

It's something we haven't yet open-sourced but use internally GitBook, let me know if you're interested and maybe I can speed up the release process.

karalabe:
I looked into a glide alternative from Rancher, called [`trash`](https://github.com/rancher/trash), which seems to easily handle everything I needed (and it's tiny at 700 LOC so I fixed the couple corner cases where it didn't work correctly and upstreamed it).

AaronO:
@karalabe Didn't know of `trash`, it looks pretty cool ! A more lightweight/pragmatic alternative to `glide`.

`glide` was the closest conceptually to `npm`, which is what the rest of our team is familiar with (since a substantial part of our code is node).

I'll add `trash.conf` support to `go-vendor-clean`.

`go-vendor-clean` is quite aggressive since it removes all non-go compilable files from `/vendor/` (`.md`, ...) along with test files (see above screenshot). Some people may want to keep that, but we wanted to keep our `/vendor/` dirs as small as possible so it works for us :)

Issue #422, opened 2016-05-15
Add option to strip *_test.go: --strip-test -t


arekkas:
Issue discussed in the Go repository: https://github.com/golang/go/issues/14417  
Godep issue: https://github.com/tools/godep/issues/427

arekkas:
Better: Make the option true by default when running > Go 1.6 or GO15VENDOREXPERIMENT

--skip-strip-tests

sdboyer:
IIRC, @mattfarina had some concerns about the licensing implications of stripping out source files. Not sure if it extended to test files, though.

Absent licensing issues, I'm +1 to this being an option. Actually, I'd argue it should be an opt in the manifest, not a CLI opt, b/c that way the behavior will be invariant across all users of the project. It's a nasty gotcha to require users to make a workflow change for this or that project when it can easily be avoided.

arekkas:
> Absent licensing issues, I'm +1 to this being an option. Actually, I'd argue it should be an opt in the manifest, not a CLI opt, b/c that way the behavior will be invariant across all users of the project. It's a nasty gotcha to require users to make a workflow change for this or that project when it can easily be avoided.

I second that.

nikhiljindal:
Came here looking for this exact feature.
There seems to be a PR for this: https://github.com/Masterminds/glide/pull/869 (last updated in July)

mattfarina:
First, if you can you should try switching to [dep](https://github.com/golang/dep).

Second, There's a plugin for what you're looking for in https://github.com/sgotti/glide-vc.

PR #869, opened 2017-07-07
Clean non source and non dependent packages

thrawn01:
This PR implements a WIP fix for #652 and #422 and is to gauge interest in this feature with the maintainers of glide. I understand new work is proceeding on [golang/dep](https://github.com/golang/dep) so work on glide may not be a priority . However, If there is interest I will continue work on this PR and make any necessary changes including adding tests, config and or cli flag options so this can be useful to the wider community 

## Purpose
To clean the vendor directory of unnecessary files and packages included in the VCS repo. Cleaning the vendor can reduce the size of a `vendor/` directory significantly. This is especially useful for projects that commit the `vendor/` directory into VCS. 

## Implementation
* When glide is provided the `--strip-vendor` flag the following is preformed
  * Remove any directory not explicit in the path of one of the declared or discovered dependencies
  * Remove any file that is not of `.go` or `.s`
  * Remove any file that has the suffix `test.go`

## Additional Work
* Add a glide.yaml config option or flag to enable cleaning `_test.go` files
* Add a glide.yaml config option or flag to enable cleaning non source files
* Add a glide.yaml config option or flag to enable cleaning all test,non source and embedded `vendor/` directories from 
* Add a glide.yaml config option to exclude a list of files and paths from cleaning
* Add tests
* During `glide init` questionnaire, ask if these options should be enabled
* Add a glide.yaml config option or flag to exclude license and legal files from cleaning
* Add a glide.yaml config option of flag to enable cleaning of files that are part of `package main` this will allow `go install ./...` to work properly with a `vendor/` directory. This is mostly useful for projects still using 1.8 or earlier as 1.9 is rumored to have this fixed. https://github.com/lparth/go/commit/eafa5c7f9b87503e63cafa64d0685e782e8f516e)

### DIFF
--- a/action/install.go
+++ b/action/install.go
@@ -63,5 +63,10 @@ func Install(installer *repo.Installer, stripVendor bool) {
 		if err != nil {
 			msg.Err("Unable to strip vendor directories: %s", err)
 		}
+		msg.Info("Cleaning test and unnecessary files from vendor directories...")
+		err = CleanVendor(newConf)
+		if err != nil {
+			msg.Err("Unable to clean vendor directories: %s", err)
+		}
 	}
 }
--- a/action/update.go
+++ b/action/update.go
@@ -2,7 +2,9 @@ package action
 
 import (
 	"io/ioutil"
+	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/Masterminds/glide/cache"
 	"github.com/Masterminds/glide/cfg"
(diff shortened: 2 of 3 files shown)


thrawn01:
Latest commit re-scans the vendored dependencies as it's unclear from looking at the Import dependencies if the root package is used by the code or simply a sub package.

`NewResolver` is not properly resolving dependencies for test files, even when specifically requesting this functionality. Looking at the code It doesn't appear to have ever worked or this might be a bug. I can't figure out how to make it resolve test dependencies and return them properly. I understand this code was not intended to be used how I'm using it so perhaps a complete re-write of `ResolveLocal()` specifically for our use case might be in order.

nikhiljindal:
It will be great to have this!
Any progress?

mattfarina:
@nikhiljindal if you need this in glide please see the [glide-vc](https://github.com/sgotti/glide-vc) plugin.

thrawn01:
Closed due to this 😢https://codeengineered.com/blog/2016/go-why-not-strip-unused-pkgs/

mattfarina:
@thrawn01 software licensing is complicated. What ever is implemented should be opt-in.

In any case, this should be discussed in dep rather than here.

PR #528, opened 2016-07-28
Build with arm versions

franciscocpg:
For this to work with `gox` this [gox PR](https://github.com/mitchellh/gox/pull/69) needs to be merged.
While it does not happen (And I don't know if someday it will), it's possible to use the fork https://github.com/franciscocpg/gox.
After merging this PR, it is necessary to run `make dist` and publish the new arm dists to the latest glide release.
Also this [glide.sh PR](https://github.com/Masterminds/glide.sh/pull/6) needs to be merged so one can install glide on a arm device with `curl https://glide.sh/get | sh`.

PR #870, opened 2017-07-10
Added support for s390x

Nayana-ibm:
Added support for  linux/s390x arch in glide. 
Please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ build-all:
 	gox -verbose \
 	-ldflags "-X main.version=${VERSION}" \
 	-os="linux darwin windows freebsd openbsd netbsd" \
-	-arch="amd64 386 armv5 armv6 armv7 arm64" \
+	-arch="amd64 386 armv5 armv6 armv7 arm64 s390x" \
 	-osarch="!darwin/arm64" \
 	-output="dist/{{.OS}}-{{.Arch}}/{{.Dir}}" .
 


Nayana-ibm:
Along with this PR, we also require to add code in gox to support s390x. 
Below diff will be required in `https://github.com/franciscocpg/gox`  repo. 

```
diff --git a/platform.go b/platform.go
index 9353b09..50820a5 100644
--- a/platform.go
+++ b/platform.go
@@ -73,6 +73,7 @@ var (
                "mips64le",
                "ppc64",
                "ppc64le",
+               "s390x",
        }

        Platforms_1_0 = []Platform{
@@ -123,7 +124,10 @@ var (
        }...)

        // Nothing changed from 1.5 to 1.6
-       Platforms_1_6 = Platforms_1_5
+       // Support for s390x
+       Platforms_1_6 = append(Platforms_1_5, []Platform{
+                {OS: "linux", Arch: "s390x", Default: true},
+       }...)
 )

 // SupportedPlatforms returns the full list of supported platforms for

```

To contribute for that change, we want to know about Contributor License Agreement.
I have checked URL to License:  https://github.com/franciscocpg/gox/blob/master/LICENSE and 
https://github.com/mitchellh/gox/blob/master/LICENSE
Please let us know which Contributor License Agreement needs for gox .

Nayana-ibm:
@mattfarina @franciscocpg Could you please provide your inputs?

franciscocpg:
@Nayana-ibm, i don't think that @mitchellh has any plans for merging new incoming PRs for https://github.com/mitchellh/gox. The last commit was about 1 year ago.

I think it makes sense to do something like [this](https://github.com/Masterminds/glide/pull/528#issuecomment-241075537).

mattfarina:
I'm going to take a look at gox vs the alternatives to see what is the best way. I'd like to cut a release soon and if there is something better now may be a good time to switch. Or, I'll look at the alternative @franciscocpg suggested.

If anyone knows of a good alternative to gox I'm interested in learning about it.

mattfarina:
I've created Masterminds/gox (based on the great work of @franciscocpg) and added support for s390x. In #916 I moved glide to use that for cross compiling. Along with this change we should now be able to produce an s390x binary.

Nayana-ibm:
Thanks @mattfarina It means Masterminds/gox will have s390x binaries/tar published on
https://github.com/Masterminds/glide/releases?

PR #878, opened 2017-07-18
`ResolveLocal(true)` now correctly reports dependencies of tests

thrawn01:
This currently does not affect any code paths, as the only usage of `ResolveLocal(true)` is `glide list` and it doesn't list test dependencies. This bug was discovered while working on #869


thrawn01:
Closing this as it's old and forgotten.

PR #883, opened 2017-07-24
resolver.ResolveLocal: fully dereference basedir before walking

drafnel:
Commit 52f7192 taught NewResolver to detect when the basedir argument is
a symlink and to extract the target of the symlink and use it in place
of the original basedir.  This works correctly and allows ResolveLocal
to walk the directory tree when the symlink target is a fully qualified
path to a directory.  But, if the symlink points to another symlink, or
if it provides a relative path, then the traversal by filepath.Walk()
will fail as before.

The path/filepath package already provides a function to fully evaluate
a path and dereference any symbolic links: filepath.EvalSymlinks().

So instead of using checkForBasedirSymlink() to perform a single
dereference of basedir and storing that into the Resolver object, let's
just record the original path into the Resolver object and then call
filepath.EvalSymlinks() to fully resolve the path in ResolveLocal before
walking the tree with filepath.Walk().

### DIFF
--- a/dependency/resolver.go
+++ b/dependency/resolver.go
@@ -185,12 +185,6 @@ func NewResolver(basedir string) (*Resolver, error) {
 		return nil, err
 	}
 
-	basedir, err = checkForBasedirSymlink(basedir)
-
-	if err != nil {
-		return nil, err
-	}
-
 	vdir := filepath.Join(basedir, "vendor")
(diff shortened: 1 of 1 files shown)


drafnel:
Ping.  Hi @mattfarina can you take a look?

Issue #873, opened 2017-07-11
[ERROR] Unable to export dependencies to vendor directory: Error moving files: exit status 1. output: Access is denied. 0 dir(s) moved.


ITSecMedia:
I've tried it 20 times, current glide doesn't work here no more on Windows 10 x64
I cleared caches, re-opened explorer, killed all handles, removed vendor ....

ITSecMedia:
[INFO]  Replacing existing vendor dependencies
... at that time glide already created **a new empty vendor folder** ...
[ERROR] Unable to export dependencies to vendor directory: Error moving files: exit status 1. output: Access is denied.
        0 dir(s) moved.

ITSecMedia:
I've replaced the following line in 

github.com/Masterminds/glide/path/**winbug.go**

**_cmd := exec.Command("cmd.exe", "/c", "copy", o, n)_**

to 

**cmd := exec.Command("robocopy.exe", o, n, "/e")**

That solved the problem, at last on my machine.

gzq0616:
func CustomRemoveAll(p string) error {

	// Handle the windows case first
	if runtime.GOOS == "windows" {
		msg.Debug("Detected Windows. Removing files using windows command")
		cmd := exec.Command("cmd.exe", "/c", fmt.Sprintf("rd /s /q %s", p))
		output, err := cmd.CombinedOutput()
		if err != nil {
			return fmt.Errorf("Error removing files: %s. output: %s", err, output)
		}
	} else if detectWsl() {
		cmd := exec.Command("rm", "-rf", p)
		output, err2 := cmd.CombinedOutput()
		msg.Debug("Detected Windows Subsystem for Linux. Removing files using subsystem command")
		if err2 != nil {
			return fmt.Errorf("Error removing files: %s. output: %s", err2, output)
		}
	}
	return os.RemoveAll(p)
}

wangxfholly:
[INFO]  --> Adding github.com/parnurzeal/gorequest to your configuration with th
e version ^0.2.15
[INFO]  Downloading dependencies. Please wait...
[INFO]  --> Fetching updates for github.com/parnurzeal/gorequest
[INFO]  Resolving imports
[INFO]  Downloading dependencies. Please wait...
[INFO]  --> Detected semantic version. Setting version for github.com/parnurzeal
/gorequest to v0.2.15
[INFO]  Exporting resolved dependencies...
[INFO]  --> Exporting github.com/parnurzeal/gorequest
[...]

PR #889, opened 2017-07-31
Fix error: Unable to export dependencies to vendor directory: Error m…

Ticore:
…oving files: exit status 1. output: Access is denied. 0 dir(s) moved.

### DIFF
--- a/path/winbug.go
+++ b/path/winbug.go
@@ -48,7 +48,7 @@ func CustomRename(o, n string) error {
 	// Handking windows cases first
 	if runtime.GOOS == "windows" {
 		msg.Debug("Detected Windows. Moving files using windows command")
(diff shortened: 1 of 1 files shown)


Ticore:
#873

Vrakfall:
Solved #873 on my end (Windows 7, no UAC but I use Comodo's HIPS instead and it didn't trigger).
Although it feels like copying many files one by one, dunno the original behaviour.

nilyang:
I have been solved this problem some days age and waiting for the PR #970 to merge to the master. 

You can also see the problem #960

prdas31:
@AndrewLader 

That's great to know! 
But from where to install the 0.12.3 version and how?
Thanks.

wuyanxin:
@prdas31  https://github.com/Masterminds/glide/releases

Issue #905, opened 2017-09-10
Glide install fails with error when spaces are in path on Windows


jpz:
```
C:\Users\Jason Zavaglia\go\src\github.com\decred\dcrd>glide install
[INFO]  Downloading dependencies. Please wait...
[INFO]  --> Found desired version locally github.com/agl/ed25519 278e1ec8e8a6e017cd07577924d6766039146ced!
...
...
[INFO]  Setting references.
[INFO]  --> Setting version for github.com/btcsuite/go-socks to 4720035b7bfd2a9bb130b1c184f8bbe41b6f0d0f.
...
...
[INFO]  Exporting resolved dependencies...
[INFO]  --> Exporting github.com/btcsuite/btclog
...
...
[INFO]  Replacing existing vendor dependencies
[ERROR] Unable to export dependencies to vendor directory: Error removing files: exit status 3. output: The system cannot find the file specified.
The system cannot find the path specified.
```

This occurs as I have GOPATH contained within my User directory, which happens to be `"FirstName LastName"`

mattfarina:
Closing as the PR has been merged.

PR #906, opened 2017-09-10
Fix CustomRemoveAll() to handle spaces in paths, and also file not found.

jpz:
Fixes to problem mentioned in https://github.com/Masterminds/glide/issues/905

This deals with two related implementation details in winbug.go -

1) the function was not working successfully, because it was not enquoting the path with spaces correctly. 

2) the exit code from Windows RD command (remove directory) should be be handled - "directory not found" should not really be an error.  By analogy, `rm -rf` returns no error code for a non-existent file.

I've tested this several times manually, but nice to get other checks on it.

### DIFF
--- a/path/path_test.go
+++ b/path/path_test.go
@@ -86,3 +86,28 @@ func TestGlide(t *testing.T) {
 	}
 	os.Chdir(wd)
 }
+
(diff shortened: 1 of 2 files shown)


mattfarina:
@jpz thanks for the contribution

PR #909, opened 2017-09-15
add --ignore to list

vetinari:
We need a package to be in GOPATH to build loadable modules outside the main repo.

A default `glide list` fails because this dependency is not in the `vendor/` tree.


szuecs:
@vetinari There is missing godoc for the ignore list and you also should provide an additional test ;)

PR #910, opened 2017-09-18
Fix issue due to go/build.ImportDir change response on not found dir

mattfarina:
In go 1.9 there was a change to the way ImportDir, or rather Import
with a local "." package, responded when not found. Previously,
the error was an OS not found error that could be detected by
os.IsNotExist. Now the error is a custom go error using the format:

    fmt.Errorf("cannot find package %q in:\n\t%s", path, p.Dir)

This change looks for both cases. For Go 1.9 is checks of it doesn't
exist because there are similar errors in go/build.

For more detail on the Go side of this see https://github.com/golang/go/issues/21923


mattfarina:
@sdboyer you might be interested in this error. Or, maybe you already have this one down.

sdboyer:
ah, thanks for the heads-up! we moved away from directly using `build.ImportDir` and rely directly on `go/parser` now, so we're set.

PR #920, opened 2017-09-29
Add recursive submodule setting for Git repositories

mspaulding06:
This PR requires a change to the `vcs` project PR I have submitted here:
https://github.com/Masterminds/vcs/pull/82/files

This is to include a new option for dependencies called `recursive` which is a boolean setting that determines if using the `--recursive` flag when working with git submodules is used or not.  Here is an example configuration file.

```yaml
package: github.com/my/package
import:
- package: gopkg.in/libgit2/git2go.v26
  recursive: false
```

I have run into a problem specifically with the `git2go` project where one of the tests has a purposefully broken `.gitmodules` file.  When the `--recursive` option is passed as the `vcs` project always does the checkout will fail and there is no way to recover.  Adding the `recursive` option allows telling Git not to deal with recursive submodules.  This is probably helpful in other scenarios as well where you are not interested in cloning recursive submodules.

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -377,6 +377,7 @@ type Dependency struct {
 	Subpackages []string `yaml:"subpackages,omitempty"`
 	Arch        []string `yaml:"arch,omitempty"`
 	Os          []string `yaml:"os,omitempty"`
+	Recursive   bool     `yaml:"recursive,omitempty"`
 }
 
 // A transitive representation of a dependency for importing and exploting to yaml.
@@ -389,6 +390,7 @@ type dep struct {
 	Subpackages []string `yaml:"subpackages,omitempty"`
 	Arch        []string `yaml:"arch,omitempty"`
 	Os          []string `yaml:"os,omitempty"`
+	Recursive   bool     `yaml:"recursive,omitempty"`
 }
 
 // DependencyFromLock converts a Lock to a Dependency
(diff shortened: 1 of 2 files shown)


mattfarina:
@mspaulding06 Have you checked this against https://github.com/golang/dep? Ideally folks will move to that and I want to make sure it's handled there as well.

/cc @sdboyer 

I'll start with the vcs PR and we can go from there.

sdboyer:
thanks for the ping. huh. interesting. yeah, no, that case hadn't occurred to me - what amounts to an intentionally broken git repository. just one more reason to hate submodules!


honestly, my first reaction is that git2go is the one breaking contracts here, and they should refactor their tests (which is the only reason I can immediately think of to have an intentionally broken .gitmodules file, but I'll go verify that assumption) to create that file on demand as part of testing, rather than having it pollute normal behavior.

mattfarina:
@sdboyer I had not considered the broken nature of the context. Great job pointing that out.

For reference libgit2/git2go#407 points out the dep, glide, and go get are all broken for it.

I'd like to see if this can be fixed upstream first. git2go is breaking contracts that are affecting a variety of package managers (and I can imagine other tools). Would be worth trying to fix it there.

mspaulding06:
I agree that this should be fixed upstream, but it still might be worth considering having an option to disable recursive submodules to avoid similar situations in the future.

Issue #432, opened 2016-05-19
Unable to resolve update issue


radamsfc:
Using glide 0.10.2

We have a glide.yaml and glide.lock generated from it.  Glide install works fine.

Updating glide.yaml in any way and trying to run an update results in errors that I have been unable to identify the source after many hours of trying to resolve this.

glide.yaml

```
package: .
import:
- package: github.com/aws/aws-sdk-go
  version: f0748927a5d7568334ff0858de06c2812da89d6a # 1.1.13
  subpackages:
  - aws
  - aws/awserr
  - aws/credentials
  - aws/session
  - service/cognitoidentity
  - service/s3
  - service/s3/s3manager
  - service/sts
- package: github.com/c9s/goprocinfo
  version: 19cb9f127a9c8d2034cf59ccb683cdb94b9deb6c
  subpackages:
  - linux
- package: github.com/coreos/etcd
  version: 2.2.5
  subpackages:
  - client
- package: github.com/davecgh/go-spew
  version: 5215b55f46b2b919f50a1df0eaa5886afe4e3b3d
  subpackages:
  - spew
- package: github.com/docker/docker
  version: 7f4e4ac6f9fd9bf46f86e6eb422cad145c75dba2
  subpackages:
  - pkg/stringutils
  - pkg/system
- package: github.com/fatih/structs
  version: 73c4e3dc02a78deaba8640d5f3a8c236ec1352bf
- package: github.com/gorilla/mux
  version: 0eeaf8392f5b04950925b8a69fe70f110fa7cbfc
- package: github.com/gorilla/websocket
  version: e2e3d8414d0fbae04004f151979f4e27c6747fe7
- package: github.com/koding/websocketproxy
  version: 870493fd19c48c3e71eaf5c5e03e07658f73bd26
- package: github.com/pmezard/go-difflib
  version: 792786c7400a136282c1664665ae0a8db921c6c2
  subpackages:
  - difflib
- package: github.com/samalba/dockerclient
  version: f274bbd0e2eb35ad1444dc6e6660214f9fbbc08c
- package: github.com/stretchr/objx
  version: 1a9d0bb9f541897e62256577b352fdbc1fb4fd94
- package: github.com/stretchr/testify/assert
  version: 6fe211e493929a8aac0469b93f28b1d0688a9a3a
- package: golang.org/x/net
  version: 6acef71eb69611914f7a30939ea9f6e194c78172
  subpackages:
  - context
  - websocket
[...]

PR #376, opened 2016-04-05
Fixes #373: Go 1.7 context package import issues

mattfarina:

PR #935, opened 2017-11-02
Fix math/bits go build tags

prateek:
This PR addresses glide's handling of go build tags for `math/bits` (added in go1.9). 

One example use I found [here](https://github.com/RoaringBitmap/roaring/blob/master/ctz.go#L1-L3)

Similar to: 
- https://github.com/Masterminds/glide/pull/376
- https://github.com/Masterminds/glide/issues/432

### DIFF
--- a/dependency/resolver.go
+++ b/dependency/resolver.go
@@ -962,6 +962,21 @@ type PkgInfo struct {
 	Loc        PkgLoc
 }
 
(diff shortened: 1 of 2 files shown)


prateek:
@mattfarina would you mind taking a look

mattfarina:
@prateek Thanks for the contribution. Time for another release with this change.

Issue #705, opened 2016-12-06
if glide is killed forcefully part way through "update" it leaves broken state


vikstrous:
Repro:
1. run glide update
2. kill -9 glide half way through
3. run glide update again

It looks like glide left a bunch of the git directories in its cache without any local files checked out and no local commits. Then when I re-ran it ran into a series of errors like the following:

```
[INFO]	--> Fetching updates for github.com/Shopify/logrus-bugsnag.
[ERROR]	Update failed for gopkg.in/yaml.v2: Unable to retrieve checked out version
[DEBUG]	Output was: HEAD
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'[DEBUG]	Unlocking https-gopkg.in-yaml.v2
```

It would be really helpful if glide could handle this state more gracefully and not error out on next update.

The manual workaround is the just remove those directories or to pull master in them.

sdboyer:
So, for starters, let's be clear - kill -9 is not something that true graceful recovery is really possible with. At least, not possible, short of designing crash-only software (the essential guarantee of which is that there is never a moment in which physical state on disk is incorrect). Achieving that standard is quite difficult, and I don't think we ever plan to have glide rise to that level of guarantee.

glide manipulates disk state, and manipulating disk state is a hard problem in general. We shell out to the individual vcs commands for various purposes; as such, we inherit all of their (potentially problematic) guarantees. 

There are weaker standards we could try to rise to - for example, defensively coding against the possibility of any bad state in the cache dirs, and trying to restore it to good state on first touch. And maybe that's something we could explore. But glide's design isn't terribly amenable to that right now, as there isn't a layer that clearly controls access to repo initialization in a way that we could hook in to. We might be able to explore something like that in [gps](https://github.com/sdboyer/gps) (which will be powering glide soon) as such a layer does exist.

(I am a big surprised about this outcome, though - signals sent via `kill`, AFAICT, do not propagate to child processes automatically, so I would've expected the git processes to run to completion when you kill the parent. I guess? not sure how this really supposed to work)

vikstrous:
Oh, haha you may be right about the kill thing. I didn't mention that I was running glide in a docker container, so it was a much more thorough killing than usual. Docker kills all subprocesses too.

[...]

PR #939, opened 2017-11-14
Change behaviour of `--force` to try a bit harder to blow away old repo if update fails for any reason

aeijdenberg:
(such as a borked cache from previously interrupted build)

Partially addresses: <https://github.com/Masterminds/glide/issues/705>

This PR splits code from the `VcsUpdate()` function into two new functions, `tryUpdate()` and `tryCheckout()` which it then invokes.

If `tryUpdate()` fails for any reason, it will trigger an attempt to delete the directory and perform `tryCheckout()` if `force` is set.

### DIFF
--- a/repo/vcs.go
+++ b/repo/vcs.go
@@ -2,6 +2,7 @@ package repo
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


aeijdenberg:
Abandoning. I think Go has moved on. :)

PR #953, opened 2017-12-12
Fix issue where error does not propogate

mattfarina:


path/strip.go
```diff
@@ -20,6 +20,10 @@ func StripVendor() error {
 	}
 
 	err := filepath.Walk(searchPath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
```
nathanraleigh-wf: Adding these lines causes an issue for me now when trying to run `glide install --strip-vendor` 

The error it produces is similar to:
[INFO]    Replacing existing vendor dependencies
[INFO]    Removing nested vendor and Godeps/_workspace directories...
[INFO]    Removing: <SOME_PATH>/lib/go/vendor
[ERROR]   Unable to strip vendor directories: open <SOME_PATH>/lib/go/vendor: no such file or directory

Do I need to create an Issue on this? It's causing our builds to fail now.

apynes-kasita: ISSUE-955 is already created

Issue #955, opened 2017-12-13
Error handling nested vendor folders


anthonyheidenreich:
Error is caused by merge of #953

```
$: glide --version
glide version 0.13.2-dev
```

glide.yaml
```
package: github.com/Kasita-Inc/tmp
import:
- package: github.com/aws/aws-sdk-go
  version: v1.10.20
```

```
$: glide --debug install --strip-vendor
[DEBUG]	No mirrors.yaml file exists
[INFO]	Downloading dependencies. Please wait...
[DEBUG]	Setting up the cache directory
[INFO]	--> Found desired version locally github.com/asaskevich/govalidator 4918b99a7cb949bb295f3c7bbaf24b577d806e35!
[INFO]	--> Found desired version locally github.com/aws/aws-sdk-go b33b08175d12b795daa7726c5577b39d06a2551c!
[INFO]	--> Found desired version locally github.com/davecgh/go-spew 04cdfd42973bb9c8589fd6a731800cf222fde1a9!
[INFO]	--> Found desired version locally github.com/docker/go-units 0dadbb0345b35ec7ef35e228dabb8de89a65bf52!
[INFO]	--> Found desired version locally github.com/go-ini/ini 32e4c1e6bc4e7d0d8451aa6b75200d19e37a536a!
[INFO]	--> Found desired version locally github.com/go-logfmt/logfmt 390ab7935ee28ec6b286364bba9b4dd6410cb3d5!
[INFO]	--> Found desired version locally github.com/go-redis/redis b879d4e0b950159c63ec69d0ed4604dcd0d97bfd!
[INFO]	--> Found desired version locally github.com/go-sql-driver/mysql a0583e0143b1624142adab07e0e97fe106d99561!
[INFO]	--> Found desired version locally github.com/go-stack/stack 259ab82a6cad3992b4e21ff5cac294ccb06474bc!
[INFO]	--> Found desired version locally github.com/golang/protobuf 1e59b77b52bf8e4b449a57e6f79f21226d571845!
[INFO]	--> Found desired version locally github.com/google/uuid 064e2069ce9c359c118179501254f67d7d37ba24!
[INFO]	--> Found desired version locally github.com/jmespath/go-jmespath dd801d4f4ce7ac746e7e7b4489d2fa600b3b096b!
[INFO]	--> Found desired version locally github.com/jmoiron/sqlx de8647470aafe4854c976707c431dbe1eb2822c6!
[...]

PR #956, opened 2017-12-13
Issue 955

apynes-kasita:
Fix for #955 
- Refactor to allow testing of the walk function
- Tests for walk function
- Return 'filepath.SkipDir' when removing a directory tree under the path as docs recommend
- Cleaning up some logic in walk function
- added integration tests that reproduce the problem


mattfarina:
Thanks for the PR. Reviewing now.

PR #974, opened 2018-02-18
Fix NPE: Check errors for WalkFunc

tamalsaha:
From https://github.com/golang/go/blob/go1.9/src/path/filepath/path.go#L336

```
// If there was a problem walking to the file or directory named by path, the
// incoming error will describe the problem and the function can decide how
// to handle that error (and Walk will not descend into that directory).
```


tamalsaha:
Here is an stacktrace for one of my Go repos that was throwing NPE:

```
```

PR #979, opened 2018-02-21
fix #977

### LINKED ISSUE #977: Make glide installable with normal `go get github.com/Masterminds/glide` command
Ay, could the glide project feature a more conventional structure, so that `go get github.com/Masterminds/glide` can install glide, just like other Go programs?

mcandre:
Adopt modern Go project structure conventions, so that glide can be installed by the much simpler `go get github.com/Masterminds/glide/...` command, like other Go programs. For local development of glide itself, `cd cmd/glide; go install` can recompile and reinstall glide.


mcandre:
Hmm, looks like AppVeyor needs updating to account for the new location of the glide CLI source?

PR #986, opened 2018-03-27
update to master if we are detached WIP

gburanov:
I have a problem 

* If the specific commit is checked out in glide, then git is in "detached head" mode, therefore pull is not working.
* Git is in state mode when we check out by tag

And because of all that the command 
```
glide up
```
is not working. We can make it work trying to checkout master


gburanov:
I understand that the fix should not be done in vendor folder, but does it make sense it all? Or am I doing smth incorrect?

PR #1033, opened 2018-10-31
Fix segfault with Glide 0.13.2 when stripping Godep workspace

databus23:
Fixes #975

When deleting 'Godeps/_workspace' folders the `Walk` function may be called for already deleted entries (e.g.  Godeps/_workspace/.gitignore) afterwards.

To reproduce the segfault try this glide.yaml:
```
package: github.com/databus23/glide-sigfault
import:
  - package: github.com/yudai/golcs
    version: 3aebbac3964febd3fa3189f5d9db214cc6e940a6
```

```
$> glide install -v
[INFO]	Lock file (glide.lock) does not exist. Performing update.
[INFO]	Downloading dependencies. Please wait...
[INFO]	--> Fetching updates for github.com/yudai/golcs
[INFO]	--> Setting version for github.com/yudai/golcs to 3aebbac3964febd3fa3189f5d9db214cc6e940a6.
[INFO]	Resolving imports
[INFO]	Downloading dependencies. Please wait...
[INFO]	Setting references for remaining imports
[INFO]	Exporting resolved dependencies...
[INFO]	--> Exporting github.com/yudai/golcs
[INFO]	Replacing existing vendor dependencies
[INFO]	Project relies on 1 dependencies.
[INFO]	Removing nested vendor and Godeps/_workspace directories...
[INFO]	Removing: /Users/d062284/Development/go/src/github.com/databus23/glide-sigfault/vendor/github.com/yudai/golcs/Godeps/_workspace
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x131c41f]

goroutine 1 [running]:
github.com/Masterminds/glide/godep/strip.stripGodepWorkspaceHandler(0xc0000f8480, 0x80, 0x0, 0x0, 0x1540480, 0xc0002da9f0, 0x0, 0x0)
	/private/tmp/glide-20180926-90535-1ihtivo/glide-0.13.2/src/github.com/Masterminds/glide/godep/strip/strip.go:62 +0x4f
path/filepath.walk(0xc00037ed80, 0x75, 0x1547760, 0xc000450b60, 0x14e0308, 0x0, 0x0)
	/usr/local/Cellar/go/1.11/libexec/src/path/filepath/path.go:378 +0x21f
path/filepath.walk(0xc000092700, 0x6a, 0x1547760, 0xc0004508f0, 0x14e0308, 0x0, 0x0)
[...]

PR #1035, opened 2018-11-07
Fix nil pointer exception

soltysh:
@mattfarina I'm not 100% why but I'm getting a nill info here, and that results in an error without that check. ptal

### DIFF
--- a/godep/strip/strip.go
+++ b/godep/strip/strip.go
@@ -58,6 +58,9 @@ func stripGodepWorkspaceHandler(path string, info os.FileInfo, err error) error 
 	if path == vPath {
 		return nil
 	}
+	if info == nil {
+		return nil
+	}
 
 	name := info.Name()
 	p := filepath.Dir(path)


godep/strip/strip.go
```diff
@@ -58,6 +58,9 @@ func stripGodepWorkspaceHandler(path string, info os.FileInfo, err error) error
 	if path == vPath {
 		return nil
 	}
+	if info == nil {
+		return nil
+	}
```
mattfarina: I wonder if the issue is also that an unhandled error is being passed as an argument into the function. That info is null because there is an error.

We ran into a similar problem in the vendor stripping handler. You can see that [here](https://github.com/Masterminds/glide/blob/9ef6ca313b3200079861f2de4e567b1d22081d19/path/strip.go#L13-L18).

Could you look to see if handling an error being passed in fixes the problem like the other function (linked) does?

soltysh: Switching to your proposed solution did not solve the problem, the root cause is `vendor/github.com/mesos/mesos-go/api/v0/examples/Godeps/_workspace/.gitignore` which is being removed earlier in the process of invoking `glide update --strip-vendor` and then leads to following error:
```
[ERROR]	Unable to strip vendor directories: lstat /data/github.com/openshift/origin/vendor/github.com/mesos/mesos-go/api/v0/examples/Godeps/_workspace/.gitignore: no such file or directory`
```

danwinship:
#1033 has the correct fix for this bug

soltysh:
Closing in favor of #1033.

Issue #1055, opened 2019-06-19
Package 


martinkunc:
Package golang.org/x/crypto/ed25519 was moved to stdlib in go 1.13
Glide is failing on resolution regardless there is a conditional build tag in the imported library.

I am importing
golang.org/x/crypto/ed25519
but glide import is failing on 
```
[DEBUG]	Trying to open golang.org/x/crypto/ed25519 (/Users/martin.kunc/.glide/cache/src/https-golang.org-x-crypto/ed25519)
[DEBUG]	Package golang.org/x/crypto/ed25519 imports crypto/ed25519
[DEBUG]	Missing crypto/ed25519. Trying to resolve.
[0;32m[INFO]	[m--> Fetching crypto/ed25519
[0;33m[WARN]	[mUnable to checkout crypto/ed25519
[0;31m[ERROR]	[mError looking for crypto/ed25519: Cannot detect VCS
```

Maybe similar to #935

flip40:
From your initial post x/crypto, you are using go version `go1.12.4`.

I believe Glide is attempting to find the built-in package `crypto/ed25519`, which you will not have since you are not yet on go1.13+. You can solve this by telling glide not to look for the package, as you know it does not exist on your current version of go. You can also likely remove this ignore statement after moving to go1.13+ and things should work normally.

Add to glide.yaml:
```
ignore:
  - crypto/ed25519
```

flip40:
Also in relation to your comment of #935, it is possible that someone needs to add it to the list of packages added to std library here:
https://github.com/Masterminds/glide/blob/master/dependency/resolver.go#L967-L978

martinkunc:
Right, thanks. Sorry for wrong linking.

PR #1056, opened 2019-06-19
crypto ed25519 added to stdlib in go1.13

martinkunc:
This should help with issue https://github.com/Masterminds/glide/issues/1055.


mattfarina:
@martinkunc thanks for the fix.

PR #1059, opened 2019-07-08
fix: git submodule foreach command accept one argument.

tangxinfa:
`glide` `update` failed on my project after upgrade git to `2.22.0`.

glide debug output:

    error: unknown switch `x'
    usage: git submodule--helper foreach [--quiet] [--recursive] [--] <command>

        -q, --quiet           Suppress output of entering each submodule command
        --recursive           Recurse into nested submodules

    fatal: run_command returned non-zero status while recursing in the nested submodules of proto

`git` `submodule` `foreach` `--recursive` command accept one argument now.

See similar merge request https://github.com/golang/dep/pull/2168/files for more information.


mattfarina:
@tangxinfa Thanks for the PR. The upstream dependency needed to be fixed instead of patching it here. That happened yesterday and a new release with a fix will come out soon.

PR #1068, opened 2020-04-07
modified the Makefile to create binary supporting power server

rchinna202:
I have add ppc64le in make file.


rchinna202:
@vibkulkarni @Nayana-ibm @mattfarina @technosophos Can you please review by PR

